PR #34, opened 2024-07-11
update peerdas impl to sync with v1.5.0-alpha5 specs

hangleang:


eip_7594/src/lib.rs
```diff
@@ -343,6 +372,64 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+    return sample_count;
+}
+
+fn hypergeom_cdf(k: u64, m: u64, n: u64, big_n: u64) -> f64 {
+    let mut sum = 0_f64;
+    for i in 0..(k + 1) {
+        sum += math_comb(n, i) * math_comb(m - n, big_n - i) / math_comb(m, big_n);
+    }
+    sum
+}
```
Karrenbelt: minor, but can use `..=`

```rust
fn hypergeom_cdf(k: u64, m: u64, n: u64, big_n: u64) -> f64 {
    let mut sum = 0_f64;
    for i in 0..=k {
        sum += math_comb(n, i) * math_comb(m - n, big_n - i) / math_comb(m, big_n);
    }
    sum
}
```

eip_7594/src/lib.rs
```diff
@@ -343,6 +372,64 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+fn math_comb(n: u64, k: u64) -> f64 {
+    if !(0..n).contains(&k) {
+        return 0_f64;
+    }
+    let mut r = 1_f64;
+    for i in 0..(std::cmp::min(k, n - k)) {
+        r *= (n - i) as f64 / (i + 1) as f64
+    }
+    r
+}
```
Karrenbelt: - the range should be 0..=n, since k can be equal to n
- handle edge cases when k is 0 or n (optional, just for sake of clarity / readability)

```rust
fn math_comb(n: u64, k: u64) -> f64 {
    if k > n {
        return 0_f64;
    }
    if k == 0 || k == n {
        return 1_f64;
    }
    let k = std::cmp::min(k, n - k);
    let mut r = 1_f64;
    for i in 0..k {
        r *= (n - i) as f64 / (i + 1) as f64;
    }
    r
}
```

types/src/eip7594.rs
```diff
@@ -16,19 +16,19 @@ use crate::{
...
-
 pub type KzgCommitmentsInclusionProofDepth = U4;
 
 pub type BlobCommitmentsInclusionProof = ContiguousVector<H256, KzgCommitmentsInclusionProofDepth>;
 
-pub const CUSTODY_REQUIREMENT: u64 = 4;
+pub const CUSTODY_REQUIREMENT: u64 = 1;
 pub const DATA_COLUMN_SIDECAR_SUBNET_COUNT: u64 = 32;
-pub const SAMPLES_PER_SLOT: u64 = 16;
+pub const SAMPLES_PER_SLOT: u64 = 8;
```
Karrenbelt: can you (refer to sources that) explain these constants?

eip_7594/src/lib.rs
```diff
@@ -343,6 +372,64 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+    let false_positive_threshold = hypergeom_cdf(
+        0,
+        NumberOfColumns::U64,
+        worst_case_missing,
+        SAMPLES_PER_SLOT,
+    );
+
+    // number of unique column IDs
+    let mut sample_count = SAMPLES_PER_SLOT;
+    while sample_count <= NumberOfColumns::U64 + 1 {
```
Karrenbelt: `while sample_count <= NumberOfColumns::U64`

eip_7594/src/lib.rs
```diff
@@ -188,47 +184,82 @@ pub fn get_custody_columns(node_id: NodeId, custody_subnet_count: u64) -> Vec<Co
...
+    let mut extended_matrix: [MatrixEntry; MAX_CELLS_IN_EXTENDED_MATRIX] =
+        core::array::from_fn(|_| MatrixEntry::default());
     for blob_index in 0..blob_count {
-        let mut cell_ids = Vec::new();
-        for &(b_index, cell_id) in cells_dict.keys() {
-            if b_index == blob_index as u64 {
-                cell_ids.push(cell_id);
+        let mut cell_indexs = vec![];
+        let mut cells = vec![];
+        let mut proofs = vec![];
```
Karrenbelt: you know the size of these and hence can be pre-allocated?

eip_7594/src/lib.rs
```diff
@@ -82,11 +80,10 @@ pub fn verify_kzg_proofs<P: Preset>(data_column_sidecar: &DataColumnSidecar<P>)
...
     );
 
-    let (row_indices, col_indices): (Vec<_>, Vec<_>) = column
+    let col_indices = column
         .iter()
-        .zip(0_u64..)
-        .map(|(_, row_index)| (row_index, index))
-        .unzip();
+        .map(|_| *index)
+        .collect::<Vec<u64>>();
```
Karrenbelt: this is just a refactor since we don't need row_indices?
`let col_indices: Vec<u64> = vec![index; column.len()];`

eip_7594/src/lib.rs
Karrenbelt: we can actually do better; handle edge cases and compute denominator only once. Moreover:
- The number of successes (i) in the sample cannot exceed the number of successes available in the population (n)
- The number of failures in the sample (big_n - i) cannot exceed the number of failures in the population (m - n)

```rust
fn hypergeom_cdf(k: u64, m: u64, n: u64, big_n: u64) -> f64 {
    if big_n > m || n > m { // Invalid input
        return 0.0;
    }
    let denominator = math_comb(m, big_n);
    let mut sum = 0.0;
    for i in 0..=k {
        if i <= n && (big_n - i) <= (m - n) {
            sum += math_comb(n, i) * math_comb(m - n, big_n - i) / denominator;
        }
    }
    sum
}

File: types/src/eip7594.rs (hunk 3 above)
Comment (hangleang) [reply types/src/eip7594.rs]: I changed based on the specs: https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/_features/eip7594/das-core.md#custody-setting

File: eip_7594/src/lib.rs (hunk 1 above)
Comment (hangleang) [reply eip_7594/src/lib.rs]: you did well calculated, I haven’t thought about that. I just try to comply the [specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/_features/eip7594/das-core.md#get_extended_sample_count), and improve the performance over time:

Comment (hangleang) [reply eip_7594/src/lib.rs]: I guess you could make a PR to update specs if you found that thing could be improved

File: eip_7594/src/lib.rs (hunk 2 above)
Comment (hangleang) [reply eip_7594/src/lib.rs]: > * the range should be 0..=n, since k can be equal to n

if `n` is included in range, so `k == n` also return 0
> * handle edge cases when k is 0 or n (optional, just for sake of clarity / readability)

found a rust util that equivalent to `math.comb` in Python: https://docs.rs/math_l/latest/src/math_l/math.rs.html#32-38
```rust
// The number of combination: k-combination of n
//  - n: a set size of n
//  - k: taken k of elements from the set
// nCk = n! / ((n - k)! * k!)
pub fn comb(n: i32, k: i32) -> i32 {
    if k > n {
        0
    } else {
        (1..=k).fold(1, |acc, val| acc * (n - val + 1) / val)
    }
}
```

eip_7594/src/lib.rs
hangleang: ahh! why I'm very bad at checking condition, lol

eip_7594/src/lib.rs
hangleang: that make sense, thank you

hangleang: that make sense, thank you

eip_7594/src/lib.rs
hangleang: looking into this

hangleang: not a constant size, but they MUST have the same size

eip_7594/src/lib.rs
```diff
@@ -343,6 +379,57 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+        if k > n {
+            0_f64
+        } else {
+            (1..=k).fold(1_f64, |acc, i| acc * (n - i + 1) as f64 / i as f64)
+        }
+    };
+    
+    let hypergeom_cdf = |k: u64, m: u64, n: u64, big_n: u64| -> f64 {
+        (0..=k).fold(0_f64, |acc, i| acc + math_comb(n, i) * math_comb(m - n, big_n - i) / math_comb(m, big_n))
+    };
```
Karrenbelt: The `math_comb` and `hypergeom_cdf` functions should be extracted out of the `get_extended_sample_count` function. Moving these to a separate module or as standalone functions will improve modularity, reusability, and testability. It will also enhance readability by separating core mathematical logic from the main function's workflow.

eip_7594/src/lib.rs
```diff
@@ -343,6 +379,57 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+ *
+ * This helper demonstrates how to calculate the number of columns to query per slot when
+ * allowing given number of failures, assuming uniform random selection without replacement.
+*/
+pub fn get_extended_sample_count(allowed_failures: u64) -> u64 {
+    // check that `allowed_failures` within the accepted range [0 -> NUMBER_OF_COLUMNS // 2]
+    assert!((0..(NumberOfColumns::U64 / 2)).contains(&allowed_failures));
+
+    // missing chunks for more than a half is the worst case
+    let worst_case_missing = NumberOfColumns::U64 / 2 + 1;
```
Karrenbelt: Is the assertion correct here? Given that `worst_case_missing` is defined as `NumberOfColumns::U64 / 2 + 1`, the assertion should match this range:

```rust
let worst_case_missing = NumberOfColumns::U64 / 2 + 1;
assert!((0..(worst_case_missing)).contains(&allowed_failures));
```

More importantly, using assert! for runtime validation can cause a panic, which isn't ideal for production code. The type checker doesn’t flag this as an issue, but a panic on failure isn’t appropriate for robust applications. Consider returning a `Result` (or `Option`) to handle errors gracefully:

```rust
pub fn get_extended_sample_count(allowed_failures: u64) -> Result<u64, &'static str> {
    let worst_case_missing = NumberOfColumns::U64 / 2 + 1;
    if !(0..(worst_case_missing)).contains(&allowed_failures) {
        return Err("allowed_failures out of range");
    }
    // remaining function logic...
}
```

eip_7594/src/lib.rs
```diff
@@ -343,6 +379,57 @@ fn kzg_commitment_inclusion_proof<P: Preset>(
...
+            worst_case_missing,
+            sample_count,
+        );
+        if prb <= false_positive_threshold {
+            break;
+        }
+        sample_count += 1;
+    }
+    return sample_count;
+}
```
Karrenbelt:
```
    (SAMPLES_PER_SLOT..=NumberOfColumns::U64)
        .find(|&count| hypergeom_cdf(
            allowed_failures,
            NumberOfColumns::U64,
            worst_case_missing,
            count,
        ) <= false_positive_threshold)
        .ok_or("Unable to find a valid sample count")
```

fork_choice_control/src/mutator.rs
```diff
@@ -1269,6 +1279,26 @@ where
...
+            self.spawn(PersistDataColumnSidecarsTask {
+                store_snapshot: self.owned_store(),
+                storage: self.storage.clone_arc(),
+                mutator_tx: self.owned_mutator_tx(),
+                wait_group,
+                metrics: self.metrics.clone(),
+            });
+        }
+    }
+
```
Karrenbelt: This implementation looks solid to me. However, while it's unrelated to your PR, I feel compelled to comment on the struct as a whole.

Is a 2500+ line struct the best approach here? It seems a bit unwieldy, and to be honest, it's raising some red flags for me in terms of maintainability. A struct isn't just a namespace, and using it in this way feels like it might be leaning too heavily on object-oriented design principles, which aren't always a natural fit in Rust.

Perhaps it's worth considering whether some of the interior mutability could be separated out from the rest of the functionality. Refactoring parts of this struct into smaller, more focused modules or components could not only make the codebase cleaner and easier to maintain but also improve performance and reduce the potential for bugs.

What do you think? Is this something we should explore further and consider raising a GitHub issue on?

p2p/src/block_sync_service.rs
```diff
@@ -266,12 +270,16 @@ impl<P: Preset> BlockSyncService<P> {
...
+                                features::log!(DebugP2p, "Batch could not retried while removing peer: {peer_id}");
+                            }
                         }
                         P2pToSync::RequestFailed(peer_id) => {
                             if !self.is_forward_synced {
                                 let batches_to_retry = self.sync_manager.remove_peer(&peer_id);
-                                self.retry_sync_batches(batches_to_retry)?;
+                                if self.retry_sync_batches(batches_to_retry).is_err() {
+                                    features::log!(DebugP2p, "Batch could not retired when request failed");
+                                }
```
Karrenbelt, line 278: :+1:

prometheus_metrics/src/metrics.rs
```diff
@@ -712,7 +738,7 @@ impl Metrics {
             tick_delay_times: GaugeVec::new(
                 opts!("TICK_DELAY_TIMES", "Tick delay times"),
                 &["tick"],
-            )?,
+            )?,
```
Karrenbelt: linter / formatter not run? There's a couple more I won't comment on since same issue

prometheus_metrics/src/metrics.rs
```diff
@@ -962,6 +995,16 @@ impl Metrics {
...
 
+    pub fn set_column_subnet_peers(&self, subnet_id: &str, num_peers: usize) {
+        self.column_subnet_peers
+            .get_metric_with_label_values(&[subnet_id])
+            .expect(
+                "the number of label values should match the number \
+                 of labels that column_subnet_peers was created with",
+            )
+            .set(num_peers as i64)
+    }
```
Karrenbelt: there was one `.expect` previously, now there is 3. Is this something worth crashing the application on? Why not log it as an error instead?

Karrenbelt:
Also, please stack PRs instead of having one PR that is as large as this (5000+ lines); that makes review a whole lot easier :pray:

prometheus_metrics/src/metrics.rs
Karrenbelt:
```
pub fn set_collection_length(&self, typename: &str, collection_name: &str, value: usize) {
    match self.collection_lengths.get_metric_with_label_values(&[typename, collection_name]) {
        Ok(metric) => metric.set(value as i64),
        Err(e) => {
            // Log the error instead of panicking
            error!(
                "Failed to set collection length for type '{}', collection '{}': {}",
                typename, collection_name, e
            );
        }
    }
}
```

types/src/eip7594.rs
Karrenbelt: thanks. Why tho is there no `TARGET_NUMBER_OF_PEERS` specified?

hangleang: not defined here, it should be in `eth2_libp2p` instead, it also got removed in the new specs

eip_7594/src/lib.rs
hangleang: good point, but i don't see it being else where. 

if we need in the future, we can seperate at that time. for now, I think it's okay as it is

fork_choice_control/src/mutator.rs
hangleang: @sauliusgrigaitis might be the right person to answer

prometheus_metrics/src/metrics.rs
hangleang: sorry, i will run `fmt` before commit, but it should be a CI for this task, isn't it?

eip_7594/src/lib.rs
hangleang: or 
```
    if allowed_failures >= worst_case_missing {
        return Err("allowed_failures out of range");
    }

PR #40, opened 2024-09-08
Tracing events

bomanaps:
Did the basic corrections, pointed out earlier.


sauliusgrigaitis:
The primary goal of the `tracing` is to have tracking of execution time of most of the functions in the client. The data should be visible in a convenient way - best would be to have that for each function there would be min time, max time, average and median time. I think it makes sense to implement this only for one module (for example `fork_choice`) just to get the feel. Please upload the numbers once you have it for running in Holesky for only one initial module. Then we can review the code and confirm that the approach is good.

bomanaps:
Thanks for the feedback! I will begin working on implementing the tracing as you suggested.

sauliusgrigaitis:
Closing in favor of #422

Issue #41, opened 2024-09-11
Windows support


mjzk:
This issue is tracking to all the problems and changes to support building Grandine on Windows.  
More about this  [Project idea](https://github.com/eth-protocol-fellows/cohort-five/blob/main/projects/grandine-windows-support.md).

Tasks:
- [x] Jemalloc building under the Windows
- [x] the dependency `psutils` does not support Windows
- [x] clear tests
- [x] test against Holesky testnet as validators
- [x] enhance current CI action with Windows support or add new Windows CI action

sauliusgrigaitis:
I'm not sure that Jemalloc is necessary on Windows. Many years ago when we had support for Windows we found that default allocator was doing very well. You can simply run some Holesky validators with Jemalloc and with the default allocator and compare memory usage patterns.

mjzk:
> I'm not sure that Jemalloc is necessary on Windows. Many years ago when we had support for Windows we found that default allocator was doing very well. You can simply run some Holesky validators with Jemalloc and with the default allocator and compare memory usage patterns.

@sauliusgrigaitis  thanks. You are right. As my testing, jemalloc-sys could be compiled with the correct building environment and some tweaks. But it is very verbose for users. Jemalloc is not well tested in Windows from the real world usage of jemalloc. Lighthouse just  disable jemalloc for Windows. So, I think it is still better to just disable it for simplicity now.

mjzk:
@sauliusgrigaitis Current PoC has been built successfully after my small metric cross platform abstraction. 

One notable change is in the [workspace Cargo.toml's lints](https://github.com/grandinetech/grandine/blob/225fb317437d33bf0392ac7b5bed24810a4f3077/Cargo.toml#L70),  I have unsafe code in new metric abstraction in that the windows crate just expose the related api in unsafe.  

I don't know if you mind this change, but currently, there's no particularly simple way for me to override lints. In fact, if you want to perform statistics like idle time on Windows, unsafe code is necessary. It's just a matter of whether you're using that unsafe by a library or in your own code.

sauliusgrigaitis:
@mjzk did you try to run Holesky validators on Windows? Let's get back to code review after the entire functionality is confirmed.

mjzk:
> @mjzk did you try to run Holesky validators on Windows? Let's get back to code review after the entire functionality is confirmed.

[...]

PR #42, opened 2024-09-12
Windows support

mjzk:
fixs #41

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,10 +4,16 @@ on: [push]
 
 jobs:
   build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     steps:
+    - name: Disable autocrlf (Windows)
+      if: runner.os == 'Windows'
+      run: git config --global core.autocrlf false
     - uses: actions/checkout@v4
       with:
         submodules: true
@@ -19,14 +25,18 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - name: Run cargo build
       run: cargo build --release --features default-networks
-    - name: Check if code is formatted
+    - name: Check if code is formatted (Linux)
+      if: runner.os == 'Linux'
       run: cargo fmt --check
-    - name: Run Clippy
+    - name: Run Clippy (Linux)
+      if: runner.os == 'Linux'
       run: scripts/ci/clippy.bash --deny warnings
     - name: Run tests
       run: cargo test --release --no-fail-fast
-    - name: Check consensus-spec-tests coverage
+    - name: Check consensus-spec-tests coverage (Linux)
+      if: runner.os == 'Linux'
       run: scripts/ci/consensus-spec-tests-coverage.rb
     # Ignore RUSTSEC-2023-0071 because we don't use RSA in `jwt-simple` 
-    - name: Run cargo audit
+    - name: Run cargo audit (Linux)
+      if: runner.os == 'Linux'
       run: cargo audit --ignore RUSTSEC-2024-0370 --ignore RUSTSEC-2023-0071
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -504,6 +504,7 @@ try_from_iterator = { path = 'try_from_iterator' }
 types = { path = 'types' }
 validator = { path = 'validator' }
 validator_key_cache = { path = 'validator_key_cache' }
+winsafe = { git = 'https://github.com/rodrigocfd/winsafe', features = ['kernel', 'psapi'] }
 
 # Banned crates
 #
--- a/ad_hoc_bench/src/main.rs
+++ b/ad_hoc_bench/src/main.rs
@@ -2,13 +2,11 @@ use core::ops::RangeInclusive;
 use std::{collections::BTreeMap, sync::Arc, time::Instant};
 
 use allocator as _;
-use anyhow::{Error, Result};
-use bytesize::ByteSize;
+use anyhow::Result;
 use clap::{Parser, ValueEnum};
 use eth2_cache_utils::{goerli, holesky, holesky_devnet, mainnet, medalla, withdrawal_devnet_4};
 use fork_choice_control::AdHocBenchController;
 use fork_choice_store::StoreConfig;
-use jemalloc_ctl::Result as JemallocResult;
 use log::info;
 use rand::seq::SliceRandom as _;
 use types::{
@@ -247,7 +245,7 @@ impl From<Blocks> for BlockParameters {
 fn main() -> Result<()> {
     binary_utils::initialize_logger(module_path!(), false)?;
     binary_utils::initialize_rayon()?;
-
+    #[cfg(not(target_os = "windows"))]
     print_jemalloc_stats()?;
 
     let options = Options::parse();
@@ -296,7 +294,7 @@ fn main() -> Result<()> {
             |_, _| BTreeMap::new(),
         ),
     }?;
-
+    #[cfg(not(target_os = "windows"))]
     print_jemalloc_stats()?;
 
     Ok(())
@@ -312,6 +310,7 @@ fn run<P: Preset>(
     beacon_blocks: impl FnOnce(RangeInclusive<Slot>, usize) -> Vec<Arc<SignedBeaconBlock<P>>>,
     blob_sidecars: impl FnOnce(RangeInclusive<Slot>, usize) -> BTreeMap<Slot, Vec<Arc<BlobSidecar<P>>>>,
 ) -> Result<()> {
+    #[cfg(not(target_os = "windows"))]
     print_jemalloc_stats()?;
 
     let Options {
@@ -422,11 +421,15 @@ fn run<P: Preset>(
     info!("average block throughput: {block_throughput:.3} blocks/s");
     info!("average slot throughput:  {slot_throughput:.3} slots/s");
 
-    print_jemalloc_stats()
+    #[cfg(not(target_os = "windows"))]
+    print_jemalloc_stats()?;
+
+    Ok(())
 }
 
+#[cfg(not(target_os = "windows"))]
 fn print_jemalloc_stats() -> Result<()> {
-    jemalloc_ctl::epoch::advance().map_err(Error::msg)?;
+    jemalloc_ctl::epoch::advance().map_err(anyhow::Error::msg)?;
 
     info!(
         "allocated: {}, \
@@ -445,9 +448,9 @@ fn print_jemalloc_stats() -> Result<()> {
 
     Ok(())
 }
-
-fn human_readable_size(result: JemallocResult<usize>) -> Result<String> {
-    let size = result.map_err(Error::msg)?;
+#[cfg(not(target_os = "windows"))]
+fn human_readable_size(result: jemalloc_ctl::Result<usize>) -> Result<String> {
+    let size = result.map_err(anyhow::Error::msg)?;
     let size = size.try_into()?;
-    Ok(ByteSize(size).to_string_as(true))
+    Ok(bytesize::ByteSize(size).to_string_as(true))
 }
--- a/metrics/src/beaconchain.rs
+++ b/metrics/src/beaconchain.rs
@@ -15,7 +15,6 @@ use helper_functions::{accessors, predicates};
 use log::warn;
 use p2p::metrics::PEERS_CONNECTED;
 use prometheus::IntGauge;
-use psutil::{cpu::CpuTimes, process::Process};
 use serde::Serialize;
 use sysinfo::{Disks, System};
 use types::{preset::Preset, traits::BeaconState};
@@ -88,29 +87,13 @@ pub struct ProcessMetrics {
 
 impl ProcessMetrics {
     pub fn get() -> Self {
-        let mut cpu_process_seconds_total = 0;
-        let mut memory_process_bytes = 0;
-
-        match Process::current() {
-            Ok(process) => {
-                match process.cpu_times() {
-                    Ok(cpu_times) => {
-                        cpu_process_seconds_total = cpu_times.busy().as_secs()
-                            + cpu_times.children_system().as_secs()
-                            + cpu_times.children_system().as_secs();
-                    }
-                    Err(error) => warn!("unable to get current process CPU usage: {error:?}"),
-                }
-
-                match process.memory_info() {
-                    Ok(mem_info) => {
-                        memory_process_bytes = mem_info.rss();
-                    }
-                    Err(error) => warn!("unable to get process memory usage: {error:?}"),
-                }
-            }
-            Err(error) => warn!("unable to get current process: {error:?}"),
-        }
+        let Ok(crate::metric_sys::ProcessCpuMetric {
+            cpu_process_seconds_total,
+            memory_process_bytes,
+        }) = crate::metric_sys::get_process_cpu_metric()
+        else {
+            panic!("unable to get current process's cpu metric");
+        };
 
         let client_build = DateTime::parse_from_rfc3339(build_time_utc!())
             .expect("unable to parse build time")
@@ -286,12 +269,16 @@ struct PlatformSpecificSystemMetrics {
 
 impl PlatformSpecificSystemMetrics {
     #[cfg(not(target_os = "linux"))]
-    fn new(_cpu: Option<&CpuTimes>) -> Self {
+    fn new() -> Self {
         Self::default()
     }
 
     #[cfg(target_os = "linux")]
-    fn new(cpu: Option<&CpuTimes>) -> Self {
+    fn new() -> Self {
+        let cpu_times = psutil::cpu::cpu_times()
+            .map_err(|error| warn!("unable to get CPU times information: {error:?}"))
+            .ok();
+        let cpu = cpu_times.as_ref();
         let mem = psutil::memory::virtual_memory()
             .map_err(|error| warn!("unable to get virtual memory information: {error:?}"))
             .ok();
@@ -325,19 +312,17 @@ impl SystemMetrics {
         let (network_node_bytes_total_receive, network_node_bytes_total_transmit) =
             helpers::get_network_bytes();
 
-        let cpu_times = psutil::cpu::cpu_times()
-            .map_err(|error| warn!("unable to get CPU times information: {error:?}"))
-            .ok();
-
-        let cpu = cpu_times.as_ref();
+        let Ok(cpu_metric) = crate::metric_sys::get_cpu_metric() else {
+            panic!("unable to get current system's cpu metric");
+        };
 
         Self {
             // CPU
             cpu_cores: system.physical_core_count().unwrap_or_default(),
             cpu_threads: system.cpus().len(),
-            cpu_node_system_seconds_total: cpu.map(CpuTimes::total).unwrap_or_default().as_secs(),
-            cpu_node_user_seconds_total: cpu.map(CpuTimes::user).unwrap_or_default().as_secs(),
-            cpu_node_idle_seconds_total: cpu.map(CpuTimes::idle).unwrap_or_default().as_secs(),
+            cpu_node_system_seconds_total: cpu_metric.system_seconds,
+            cpu_node_user_seconds_total: cpu_metric.user_seconds,
+            cpu_node_idle_seconds_total: cpu_metric.idle_seconds,
 
             // memory
             memory_node_bytes_total: system.total_memory(),
@@ -359,7 +344,7 @@ impl SystemMetrics {
             misc_os: metrics_os(),
 
             // platform specific metrics
-            platform_specific_metrics: PlatformSpecificSystemMetrics::new(cpu),
+            platform_specific_metrics: PlatformSpecificSystemMetrics::new(),
         }
     }
 }
--- a/metrics/src/metric_sys.rs
+++ b/metrics/src/metric_sys.rs
@@ -1,0 +1,123 @@
+use anyhow::Result;
+#[cfg(target_os = "linux")]
+use {
+    anyhow::bail,
+    psutil::{cpu, process::Process},
+};
+#[cfg(target_os = "windows")]
+use {
+    log::debug,
+    winsafe::{self as w, prelude::*},
+};
+
+pub struct ProcessCpuMetric {
+    pub cpu_process_seconds_total: u64,
+    pub memory_process_bytes: u64,
+}
+
+#[cfg(target_os = "linux")]
+pub fn get_process_cpu_metric() -> Result<ProcessCpuMetric> {
+    #[allow(unused_assignments)]
+    let mut cpu_process_seconds_total = 0;
+    #[allow(unused_assignments)]
+    let mut memory_process_bytes = 0;
+    match Process::current() {
+        Ok(process) => {
+            match process.cpu_times() {
+                Ok(cpu_times) => {
+                    cpu_process_seconds_total = cpu_times.busy().as_secs()
+                        + cpu_times.children_system().as_secs()
+                        + cpu_times.children_system().as_secs();
+                }
+                Err(error) => bail!("unable to get current process CPU usage: {error:?}"),
+            }
+
+            match process.memory_info() {
+                Ok(mem_info) => {
+                    memory_process_bytes = mem_info.rss();
+                }
+                Err(error) => bail!("unable to get process memory usage: {error:?}"),
+            }
+        }
+        Err(error) => bail!("unable to get current process: {error:?}"),
+    }
+    Ok(ProcessCpuMetric {
+        cpu_process_seconds_total,
+        memory_process_bytes,
+    })
+}
+
+#[allow(unused_assignments)]
+#[cfg(target_os = "windows")]
+pub fn get_process_cpu_metric() -> Result<ProcessCpuMetric> {
+    let proc = w::HPROCESS::GetCurrentProcess();
+
+    // Get CPU times
+    let (_, _, kernel, user) = proc.GetProcessTimes()?;
+    let kernel_seconds = filetime_to_seconds(kernel);
+    let user_seconds = filetime_to_seconds(user);
+    let cpu_process_seconds_total = kernel_seconds + user_seconds;
+    debug!("CPU time: {:.2} seconds", cpu_process_seconds_total);
+
+    // Get memory info
+    let mem_info = proc.GetProcessMemoryInfo()?;
+    let memory_process_bytes = mem_info.WorkingSetSize.try_into()?;
+    debug!("memory usage: {} bytes", memory_process_bytes);
+
+    Ok(ProcessCpuMetric {
+        cpu_process_seconds_total,
+        memory_process_bytes,
+    })
+}
+
+#[derive(Debug)]
+pub struct CpuMetric {
+    pub idle_seconds: u64,
+    pub system_seconds: u64,
+    pub user_seconds: u64,
+}
+
+// TODO maybe work for MacOS or wider Unix?
+#[cfg(target_os = "linux")]
+pub fn get_cpu_metric() -> Result<CpuMetric> {
+    let cpu = cpu::cpu_times()?;
+    let system_seconds = cpu.total().as_secs();
+    let user_seconds = cpu.user().as_secs();
+    let idle_seconds = cpu.idle().as_secs();
+    Ok(CpuMetric {
+        idle_seconds,
+        system_seconds,
+        user_seconds,
+    })
+}
+
+#[cfg(target_os = "windows")]
+pub fn get_cpu_metric() -> Result<CpuMetric> {
+    let (idle_time, kernel_time, user_time) = w::GetSystemTimes()?;
+
+    // Convert FILETIME to u64 (100-nanosecond intervals)
+    let idle = filetime_to_seconds(idle_time);
+    let kernel = filetime_to_seconds(kernel_time);
+    let user = filetime_to_seconds(user_time);
+
+    // Calculate system time (kernel time includes idle time)
+    let system = kernel - idle;
+
+    Ok(CpuMetric {
+        idle_seconds: idle,
+        system_seconds: system,
+        user_seconds: user,
+    })
+}
+
+// NOTE:
+//   FILETIME: Contains a 64-bit value representing the number of 100-nanosecond intervals since January 1, 1601 (UTC). (Ref: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime)
+//   But some Windows APIs just use it to represent a relative time interval, a.k.a., duration.
+//   For example, GetSystemTimes like APIs are used in this mod.
+//   The helper function just converts the number of 100-nanosecond into the number of seconds
+#[inline(always)]
+#[cfg(target_os = "windows")]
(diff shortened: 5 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -1,27 +1,33 @@
...
+      if: runner.os == 'Windows'
+      run: git config --global core.autocrlf false
     - uses: actions/checkout@v4
       with:
         submodules: true
         lfs: true
-    # There should be no need to install Rustup or a Rust toolchain explicitly.
-    # The `ubuntu-22.04` image already includes Rustup:
-    # <https://github.com/actions/runner-images/blob/ubuntu22/20240401.4/images/ubuntu/Ubuntu2204-Readme.md#rust-tools>
-    # Running `rustc` or Cargo should automatically install the toolchain specified in `rust-toolchain.toml`.
```
weekday-grandine-io: [**suggestion**](https://conventionalcomments.org/): Instead of removing the comment, update it to explain the same is true for `windows-2022`.

```yaml
    # The `ubuntu-22.04` and `windows-2022` images already include Rustup:
    # - <https://github.com/actions/runner-images/blob/ubuntu22/20240401.4/images/ubuntu/Ubuntu2204-Readme.md#rust-tools>
    # - <https://github.com/actions/runner-images/blob/ubuntu22/20240401.4/images/windows/Windows2022-Readme.md#rust-tools>
```

.github/workflows/ci.yml
```diff
@@ -1,27 +1,33 @@
...
-    - name: Check if code is formatted
+    - name: Check if code is formatted (Linux)
+      if: runner.os == 'Linux'
       run: cargo fmt --check
-    - name: Run Clippy
-      run: scripts/ci/clippy.bash --deny warnings
+    - name: Run Clippy (Linux)
+      if: runner.os == 'Linux'
+      shell: bash
+      run: bash scripts/ci/clippy.bash --deny warnings
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): Specifying Bash as the interpreter explicitly should not be necessary.

This step will only run on Linux.
It has worked on Linux until now.

I assume the addition of `bash` is an oversight.

```suggestion
      run: scripts/ci/clippy.bash --deny warnings
```

Cargo.toml
```diff
@@ -67,7 +67,7 @@ resolver = '2'
 edition = '2021'
 
 [workspace.lints.rust]
-unsafe_code = 'forbid'
+unsafe_code = 'deny'
```
weekday-grandine-io: [**issue**](https://conventionalcomments.org/): I am opposed to the use of `unsafe` in this workspace.

Try using [`winsafe`].
[`winsafe`] does not have a wrapper for [`GetProcessMemoryInfo`], but you could fork it and even submit a pull request [upstream].

Also, it makes little sense to set a lint like [`unsafe_code`] to `deny`.
Unsafe Rust is already opt-in via the `unsafe` keyword.

[`winsafe`]: https://crates.io/crates/winsafe
[`GetProcessMemoryInfo`]: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/ProcessStatus/fn.GetProcessMemoryInfo.html
[upstream]: https://github.com/rodrigocfd/winsafe
[`unsafe_code`]: https://doc.rust-lang.org/1.80.1/rustc/lints/listing/allowed-by-default.html#unsafe-code

Cargo.toml
```diff
@@ -497,6 +497,12 @@ try_from_iterator = { path = 'try_from_iterator' }
 types = { path = 'types' }
 validator = { path = 'validator' }
 validator_key_cache = { path = 'validator_key_cache' }
+windows = { version = "0.58", features = [
+    "Win32_Foundation",
+    "Win32_System_ProcessStatus",
+    "Win32_System_Threading",
+    "Win32_System_SystemInformation",
+] }
```
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): We have a convention of using [literal strings] in TOML files [whenever possible].

Also, we usually order features alphabeticaly and do not wrap them over multiple lines.

```suggestion
windows = { version = '0.58', features = [
    'Win32_Foundation',
    'Win32_System_ProcessStatus',
    'Win32_System_SystemInformation',
    'Win32_System_Threading',
] }
```
or
```suggestion
windows = { version = '0.58', features = ['Win32_Foundation', 'Win32_System_ProcessStatus', 'Win32_System_SystemInformation', 'Win32_System_Threading'] }
```

[literal strings]: https://toml.io/en/v0.3.0#string
[whenever possible]: https://en.wikipedia.org/wiki/Rule_of_least_power

ad_hoc_bench/src/main.rs
```diff
@@ -422,11 +421,14 @@ fn run<P: Preset>(
...
-    print_jemalloc_stats()
+    #[cfg(not(target_os = "windows"))]
+    print_jemalloc_stats()?;
+    Ok(())
 }
 
+#[cfg(not(target_os = "windows"))]
 fn print_jemalloc_stats() -> Result<()> {
-    jemalloc_ctl::epoch::advance().map_err(Error::msg)?;
+    jemalloc_ctl::epoch::advance().map_err(anyhow::Error::msg)?;
```
weekday-grandine-io, line 432: [**nitpick**](https://conventionalcomments.org/): Import types at the top of the file.

```rust
#[cfg(target_os = "windows")]
use ::{anyhow::Error, bytesize::ByteSize, jemalloc_ctl::Result as JemallocResult};
```

ad_hoc_bench/src/main.rs
```diff
@@ -422,11 +421,14 @@ fn run<P: Preset>(
     info!("average block throughput: {block_throughput:.3} blocks/s");
     info!("average slot throughput:  {slot_throughput:.3} slots/s");
 
-    print_jemalloc_stats()
+    #[cfg(not(target_os = "windows"))]
+    print_jemalloc_stats()?;
+    Ok(())
```
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): Add an empty line before `Ok(())`.

That would match the prevailing style in our codebase.

ad_hoc_bench/src/main.rs
```diff
@@ -2,13 +2,11 @@ use core::ops::RangeInclusive;
 use std::{collections::BTreeMap, sync::Arc, time::Instant};
 
 use allocator as _;
-use anyhow::{Error, Result};
-use bytesize::ByteSize;
+use anyhow::{Ok, Result};
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): There is usually no need to import `anyhow::Ok`.

The return type of `run` should provide enough information to infer the type parameter.

metrics/src/beaconchain.rs
```diff
@@ -15,12 +15,15 @@ use helper_functions::{accessors, predicates};
...
 use prometheus::IntGauge;
-use psutil::{cpu::CpuTimes, process::Process};
 use serde::Serialize;
 use sysinfo::{Disks, System};
 use types::{preset::Preset, traits::BeaconState};
 
-use crate::{helpers, MetricsService};
+use crate::{
+    helpers,
+    metric_sys::{get_cpu_metric, get_process_cpu_metric},
```
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): Module-qualify or crate-qualify functions instead of importing them.

For example, `get_cpu_metric` should be referred to as `metric_sys::get_cpu_metric` or `crate::metric_sys::get_cpu_metric`.

We follow [this ancient style guide].

[this ancient style guide]: https://doc.rust-lang.org/1.0.0/style/style/imports.html#prefer-fully-importing-types/traits-while-module-qualifying-functions.

metrics/src/beaconchain.rs
```diff
@@ -88,29 +91,13 @@ pub struct ProcessMetrics {
...
-            }
-            Err(error) => warn!("unable to get current process: {error:?}"),
-        }
+        let Ok(crate::metric_sys::ProcessCpuMetric {
+            cpu_process_seconds_total,
+            memory_process_bytes,
+        }) = get_process_cpu_metric()
+        else {
+            panic!("unable to get current process's cpu metric");
+        };
```
weekday-grandine-io: [**issue**](https://conventionalcomments.org/): Propagate or handle the error instead of panicking.

Propagating it should be the easier option:
```suggestion
        }) = get_process_cpu_metric()?;
```
Change the return type of `ProcessMetrics::get` if needed.

The same applies to the `panic!` in `SystemMetrics::get`.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+#[cfg(target_os = "windows")]
+use std::mem::size_of;
+#[cfg(target_os = "windows")]
+use windows::Win32::{
+    Foundation::FILETIME,
+    System::{
+        ProcessStatus::{GetProcessMemoryInfo, PROCESS_MEMORY_COUNTERS},
+        Threading::{GetCurrentProcess, GetProcessTimes, GetSystemTimes},
+    },
+};
```
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): Group conditional imports.

```suggestion
use anyhow::{Ok, Result};

#[cfg(target_os = "linux")]
use ::{
    log::warn,
    psutil::cpu::{self, CpuTimes},
};

#[cfg(target_os = "windows")]
use ::{
    anyhow::bail,
    log::trace,
    std::mem::size_of,
    windows::Win32::{
        Foundation::FILETIME,
        System::{
            ProcessStatus::{GetProcessMemoryInfo, PROCESS_MEMORY_COUNTERS},
            Threading::{GetCurrentProcess, GetProcessTimes, GetSystemTimes},
        },
    },
};
```

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+use anyhow::bail;
+use anyhow::{Ok, Result};
+#[cfg(target_os = "windows")]
+use log::trace;
+#[cfg(target_os = "linux")]
+use log::warn;
+#[cfg(target_os = "linux")]
+use psutil::cpu::{self, CpuTimes};
+#[cfg(target_os = "windows")]
+use std::mem::size_of;
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): Importing `size_of` should not be necessary on any platform.

`size_of` is in the prelude starting with Rust 1.80.0.

If it weren't, we would call it as `core::mem::size_of` instead of importing it.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+pub fn get_process_cpu_metric() -> Result<ProcessCpuMetric> {
+    #[cfg(target_os = "linux")]
+    {
+        get_process_cpu_metric_linux()
+    }
+    #[cfg(target_os = "windows")]
+    {
+        get_process_cpu_metric_win()
+    }
+}
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): `get_process_cpu_metric` serves no purpose.

Remove `get_process_cpu_metric`.
Rename `get_process_cpu_metric_linux` and `get_process_cpu_metric_win` to `get_process_cpu_metric`.
They can have the same name because no more than one of them will be included in the build.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+
+#[cfg(target_os = "linux")]
+fn get_process_cpu_metric_linux() -> Result<ProcessCpuMetric> {
+    use anyhow::bail;
+    use core::result::Result::Ok;
+    use psutil::process::Process;
```
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): Import types at the top of the file.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[allow(unused_assignments)]
+    let mut cpu_process_seconds_total = 0;
+    #[allow(unused_assignments)]
+    let mut memory_process_bytes = 0;
```
weekday-grandine-io, line 23: [**polish**](https://conventionalcomments.org/): Leave the variables uninitialized to pass the `unused_assignments` lint.

```suggestion
    let cpu_process_seconds_total;
    let memory_process_bytes;
```

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+        cpu_process_seconds_total,
+        memory_process_bytes,
+    })
+}
+
+#[allow(unsafe_code, unused_assignments)]
+#[cfg(target_os = "windows")]
+fn get_process_cpu_metric_win() -> Result<ProcessCpuMetric> {
+    let mut cpu_process_seconds_total = 0;
+    let mut memory_process_bytes: u64 = 0;
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): Leave the variables uninitialized to pass the `unused_assignments` lint.

```suggestion
    let cpu_process_seconds_total;
    let memory_process_bytes;
```

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    if unsafe {
+        GetProcessMemoryInfo(
+            process_handle,
+            &mut mem_counters,
+            size_of::<PROCESS_MEMORY_COUNTERS>() as u32,
+        )
+        .is_ok()
+    } {
+        memory_process_bytes = mem_counters.WorkingSetSize as _;
```
weekday-grandine-io: [**issue**](https://conventionalcomments.org/): This seems to cast `usize` to `u64`.

Casting can silently truncate values.
The issue is mostly theoretical for this combination of types, but it's easy to avoid.
Use [`TryInto::try_into`] instead of casting:
```suggestion
        memory_process_bytes = mem_counters.WorkingSetSize.try_into()?;
```

[`TryInto::try_into`]: https://doc.rust-lang.org/stable/core/convert/trait.TryInto.html#tymethod.try_into

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+        trace!("Memory usage: {} bytes", memory_process_bytes);
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): Use argument capture.

```suggestion
        trace!("Memory usage: {memory_process_bytes} bytes");
```

weekday-grandine-io: [**suggestion (non-blocking)**](https://conventionalcomments.org/): Consider using `debug!`.

We rarely use `trace!`.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+        )
+        .is_ok()
+    } {
+        let kernel_seconds = filetime_to_seconds(&kernel_time);
+        let user_seconds = filetime_to_seconds(&user_time);
+        cpu_process_seconds_total = kernel_seconds + user_seconds;
+        trace!("CPU time: {:.2} seconds", cpu_process_seconds_total);
+    } else {
+        bail!("Unable to get process CPU usage");
+    }
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): Instead of returning a new error, propagate the one returned by `GetProcessTimes` and add context using [`anyhow::Context::context`].

```suggestion
    unsafe {
        GetProcessTimes(
            process_handle,
            &mut creation_time,
            &mut exit_time,
            &mut kernel_time,
            &mut user_time,
        )
        .context("unable to get process CPU usage")?;
    }
    
    let kernel_seconds = filetime_to_seconds(&kernel_time);
    let user_seconds = filetime_to_seconds(&user_time);
    cpu_process_seconds_total = kernel_seconds + user_seconds;
    
    trace!("CPU time: {cpu_process_seconds_total:.2} seconds");
```

[`anyhow::Context::context`]: https://docs.rs/anyhow/1.0.89/anyhow/trait.Context.html#tymethod.context

metrics/src/metric_sys.rs
weekday-grandine-io: [**nitpick**](https://conventionalcomments.org/): Most error and log messages in the codebase start in lowercase.

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+    Ok(CpuMetric {
+        idle_seconds: idle,
+        system_seconds: system,
+        user_seconds: user,
+    })
+}
+
+#[inline(always)]
+#[cfg(target_os = "windows")]
+fn filetime_to_seconds(ft: &FILETIME) -> u64 {
```
weekday-grandine-io: [**polish**](https://conventionalcomments.org/): It would be more idiomatic to pass [`FILETIME`] as an owned value.

[`FILETIME`] is a small struct that implements `Copy`.

```suggestion
fn filetime_to_seconds(ft: FILETIME) -> u64 {
```

[`FILETIME`]: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/struct.FILETIME.html

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+    let total_seconds = ((ft.dwHighDateTime as u64) << 32 | ft.dwLowDateTime as u64) / 10_000_000;
+    total_seconds
+}
```
weekday-grandine-io: [**thought**](https://conventionalcomments.org/): It would be more idiomatic to return [`Duration`], but that's probably out of scope.

Representing the number of seconds with `u64` limits precision to 1 second, but that seems to be required by the [Client Stats Proposal].

[`Duration`]: https://doc.rust-lang.org/stable/core/time/struct.Duration.html
[Client Stats Proposal]: https://docs.google.com/document/d/1qPWAVRjPCENlyAjUBwGkHMvz9qLdd_6u9DPZcNxDBpc

metrics/src/metric_sys.rs
```diff
@@ -0,0 +1,203 @@
...
+        idle_seconds: idle,
+        system_seconds: system,
+        user_seconds: user,
+    })
+}
+
+#[inline(always)]
+#[cfg(target_os = "windows")]
+fn filetime_to_seconds(ft: &FILETIME) -> u64 {
+    let total_seconds = ((ft.dwHighDateTime as u64) << 32 | ft.dwLowDateTime as u64) / 10_000_000;
```
weekday-grandine-io, line 121: [**suggestion**](https://conventionalcomments.org/): Document the magic number somehow.

Consider extracting a constant with a self-documenting name.
Consider linking to the documentation of [`FILETIME`].

[`FILETIME`]: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime

weekday-grandine-io, line 121: [**note**](https://conventionalcomments.org/): [`FILETIME`] is a kind of timestamp (with an unusual epoch), but [`GetSystemTimes`] uses it to represent a duration.

[`GetSystemTimes`]: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getsystemtimes
[`FILETIME`]: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime

weekday-grandine-io (changes requested):

mjzk:
@weekday-grandine-io  thanks for the detailed review. I will react according to my understanding. Although I don't think some of your opinions are necessarily valid, I will give priority to your suggestions.

.github/workflows/ci.yml
mjzk: OK. My personal think for the removal here is: These comments are of little significance. `ubuntu-latest` is not `ubuntu-22.04` mentioned here but [`Ubuntu 24.04`](https://github.com/actions/runner-images?tab=readme-ov-file#available-images). It doesn't make much sense to let developers maintain these changes.

But, I can also accept that if you, as a committer, think it is necessary to keep. So I will modify it according to your suggestion.

.github/workflows/ci.yml
mjzk: OK.  The addition of bash is intentional. The two run command diff here is: the original requires the script `scripts/ci/clippy.bash` in the executable mode. This first has security concerns. But most problem here is, if you check out the script in unsupport OSes and FSes like in Windows, the executable mode will be lost. Then, although you want to execute it in WSL like things. It does not work. 

But, I can also accept that if you, as a committer, think you want to use the original command. So I will modify it according to your suggestion.

Cargo.toml
mjzk: @weekday-grandine-io Basically, I do not agree with you on this point. 

1. [The sense of `deny`](https://doc.rust-lang.org/rustc/lints/levels.html#deny) is to make developer have the chance to opt-out forced lint errors. Here, the result of the behavior that I override that unsafe lint error is controllable and has minimal impact. To forbid unsafe in a large Rust engineering, like Grandine, is just a cover-up in the current situation where we heavily depends on a large number of unsafe libraries.

2. Although firing a pull request [upstream](https://github.com/rodrigocfd/winsafe) is possible, it is time-consuming, and no guarantee of upstream willing to accept new additions or modifications. I think, that's why we have so many forks in our  grandinetech org.

Although I can not agree with you on this point, I can also accept that if you, as a committer, think you just want to use `unsafe_code = 'forbid'`. But I suggest to reconsidering the fixing way.

There are three kinds of actions now:
1.  keep this change in PR and nothing need to be done.
2. firstly to fork a project in our grandinetech org, to quick apply update in current PR. Then, make Grandine depends on it and go on...
3. firstly to fork a project anywhere, to submit an PR upstream. Then, to wait it accepted and make Grandine depends on it after that and go on...

@weekday-grandine-io @sauliusgrigaitis What do you think?

Cargo.toml
mjzk: accepted.

ad_hoc_bench/src/main.rs
mjzk: This import is just used for `not(target_os = "windows")`. To prompt it to the top in import will cause additional verbose imports. 

But, I can also accept that if you, as a committer, think you want to use this more verbose imports.

ad_hoc_bench/src/main.rs
mjzk: accepted.

ad_hoc_bench/src/main.rs
mjzk: accepted. It is interesting to see that our clippy check takes no effect for this unsed import.

metrics/src/beaconchain.rs
mjzk: accepted. 

Interesting. But I think there may be many places which does not adhere to your said style like [this get function usage](https://github.com/mjzk/grandine/blob/windows_support/metrics/src/server.rs#L14) (just for example).

metrics/src/beaconchain.rs
mjzk: accepted. I will use the original `warn!` way although it is more stupid in my thinking.

Propagating is not an easier option otherwise the original code just do it. And the entire codebase has a lot of panics, I just follow that existing style and do not change any bad behaviors to the whole project. I personally think you have double standards. 😺

Cargo.toml
sauliusgrigaitis: So in the core team, we think that `unsafe` code is really not desired in Grandine codebase (dependencies is another story). So in order to ensure this philosophy we use `unsafe_code = 'forbid'`. For sure it's not very convenient whenever a tricky situation comes, but we try to be creative to workaround particular cases so we can still proceed using  `unsafe_code = 'forbid'`.

mjzk: @sauliusgrigaitis OK. 

I understand our decision although I think it is not very meaningful now. For example, our bad stackoverflow in the debug mod currently. The philosophy of `deny`, in fact, also forbids the unsafe by default. There is no unintentional unsafe. But again, I respect the core team's decision.

I will follow the above  action#3 which was suggested by @weekday-grandine-io , although you didn't suggest which is better in 2 and 3.

Apparently this causes an issue that this PR is being locked by upstream. I will research and contact upstream as soon as possible to resolve this issue. However, there may not be a clear timeline.

metrics/src/metric_sys.rs
mjzk: Yes, it is better to group conditional imports.  The main reason to use separated imports is to facilitate testing in that time.

But, for group conditional imports, "::" is unnecessary in fact.

metrics/src/metric_sys.rs
mjzk: gone with new impl based on winsafe

metrics/src/metric_sys.rs
mjzk: accepted. 

This follow my own conventions. Conditional compilation is not always mutually exclusive. A consistent style is better for reading.  YMMV.

metrics/src/metric_sys.rs
mjzk: accepted.

metrics/src/metric_sys.rs
mjzk: gone with new impl based on winsafe

metrics/src/metric_sys.rs
mjzk: gone with new impl based on winsafe

metrics/src/metric_sys.rs
mjzk: gone with new impl based on winsafe.

But I don't agree with your statement. Firstly, the size of `usize` is not larger than that of `u64`. There is no silently truncation here. Secondly, in most case, the safety of casting is damn easy to be guarded by coders.  Universal `TryInto::try_into` is not a good idea. It forces the introduction of error handling. It is verbose for most cases and performance forbidden in some cases. Coding standards are not a dead standard, we need to balance it according to the situation.

metrics/src/metric_sys.rs
mjzk: accepted.

But my personally think this style is a bit excessive

mjzk: accepted.

metrics/src/metric_sys.rs
mjzk: gone with new impl based on winsafe.

metrics/src/metric_sys.rs
mjzk: accepted.

metrics/src/metric_sys.rs
mjzk: accepted.

(But passing by ref/pointer indeed follows the FFI convention in the Rust community.)

metrics/src/metric_sys.rs
mjzk: It is interesting to see that we have a specification for client stats. This method is indeed a quick method to imitate the original Linux side logic. We can change it in the future if necessary.

metrics/src/metric_sys.rs
mjzk: accepted.

mjzk: ok.

mjzk:
@sauliusgrigaitis @weekday-grandine-io 

For syncing PR to `develop` branch easier, I do a squash on all commits. And finally [CI passed again](https://github.com/mjzk/grandine/actions/runs/11699850295). 

After contacting with upstream, the relevant Windows stats successfully rely on winsafe. This also greatly simplifies the implementation of the metric_sys mod.

Other changes are also made according to the review comments even with some of the changes which I not fully agree with. 

If you need any other changes, please let me know and I will try to complete them even after the end of EPF cohort 5.

metrics/src/beaconchain.rs
weekday-grandine-io: The code in the diff I see still seems to be using `panic!`.

> the entire codebase has a lot of panics

The rule we follow (or try to) is to not panic in response to conditions at runtime.
The use of panicking macros and methods like `Result::except` is permitted if the developer believes that the failure case cannot occur at runtime (for example, when making assertions about invariants).
In other words, a panic at runtime should always indicate a bug.
Conditions that can occur in a correctly functioning application should be handled without panics.
`Result` is a common way to achieve that.

metrics/src/metric_sys.rs
weekday-grandine-io: > Conditional compilation is not always mutually exclusive.

Can `target_os = "linux"` and `target_os = "windows"` be true at the same time?
As far as I know they can't.

If both predicates can be true at the same time, then the original implementation (the one I commented on) was incorrect.
The call to `get_process_cpu_metric_linux` would trigger a compilation error if both predicates were true.

metrics/src/metric_sys.rs
weekday-grandine-io: They are still present in the diff I see.
You may have been thinking of the thread below.

metrics/src/metric_sys.rs
weekday-grandine-io: > Firstly, the size of usize is not larger than that of u64.

True in practice but not in theory.
There are platforms with a word size greater than 64 bits.
The EVM is a notable one.
I don't think Rust supports any such platform.

weekday-grandine-io:
I don't see any serious problems, but there are small ones that I believe can be easily resolved.
See the other 4 comments I posted.
I'd prefer if they were resolved, but I'll defer to @sauliusgrigaitis.

metrics/src/metric_sys.rs
mjzk: @weekday-grandine-io No, they can't. What I meant to say is that this is a style of conditional compilation. It is more friendly to non-mutually exclusive conditions. 

If both predicates can be true at the same time, then the original implementation (the one I coded and your commented on) was **_not necessarily incorrect_**.

If we just ignore all the detailed _get_process_cpu_metric_*_, they are just two calls. Two calls are allowed to run freely according to the real conditions. 

The essence here is that with the style I use, there are no naming conflicts in the code, but your style exploits naming conflicts. But naming conflicts will confuse some people who don't use IDE to read the code. OK, this is just my 2 cent.

metrics/src/beaconchain.rs
mjzk: @weekday-grandine-io 

> The use of panicking macros and methods like Result::except is permitted if the developer believes that the failure case cannot occur at runtime 

Your point of view is actually specious. Bugs or poorly thought-out code can invalidate invariants. Without a review of the entire system, who can determine whether a branching is for the invariant or just as an ordinary condition?

The best practice is to completely disable panic unless you want fail-fast.

For these codes, I actually forgot to change this code. If throwing errors here, a lot of code needs to be changed. I have a headache here. The original logic is to use `unwrap_or_default`. This implementation is really not good because you can't distinguish whether it is an error or a true zero value.

If you are OK, I want to use follow the idea of `unwrap_or_default`. That's if error, we just use the default value (zero here). This makes life easier.  I will change this in the weekend.

metrics/src/metric_sys.rs
mjzk: @weekday-grandine-io Yes, I have changed this in latest commit. 

But is this an overreaction? Although EVM like VM may provide larger word size, I just discuss the practical clean and performance-wise codes. The problem also lies in Ethereum: now we know the quantum processor can compromise Ethereum, then whether the Ethereum protocol need to be modified immediately to resist quantum attacks?

Tumas:
Hi, 

I've rebased your branch on latest develop, solved the conflicts and merged this manually to keep the history clean:  0a589082cf535c8952e07dc47e0941a7df1f5cb6

mjzk:
@Tumas Thanks for your rebasing and merging! And also thanks for all the helps in this issue, including @weekday-grandine-io 's detailed reviewing and @sauliusgrigaitis 's guiding in this EPF. See you in the community and in the future!

PR #54, opened 2024-10-30
fix: duplicate get_attestation_participation_flags for Deneb phase to streamline participation flag logic

Kayden-ML:
Changes Made:

Refactored Functionality: Introduced get_attestation_participation_flags_deneb to handle participation flag logic specific to the Deneb phase. This separation of logic ensures that the Deneb functionality is clearly defined and distinct from earlier phases.

Eliminated Conditional Complexity: By removing the need to check the state phase within the original function, the code is now cleaner and more straightforward. This reduces cognitive overhead when reading and maintaining the code.

### DIFF
--- a/helper_functions/src/accessors.rs
+++ b/helper_functions/src/accessors.rs
@@ -677,9 +677,48 @@ pub fn get_attestation_participation_flags<P: Preset>(
         participation_flags.set_bit(TIMELY_SOURCE_FLAG_INDEX, true);
     }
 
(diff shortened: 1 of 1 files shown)


povi:
Hi,

Looks like `get_active_balance` is not used anywhere?

Also, a modified `get_attestation_participation_flags_deneb` function is called at the end of the original function, so basically the code in original function is executed at runtime without any benefit before a modified one is called.

Finally, it should be `deneb::get_attestation_participation_flags` and not `get_attestation_participation_flags_deneb`, same as modified functions in `helper_functions` are.

PR #58, opened 2024-12-11
WIP: Fulu fork

hangleang:
extend phase with `fulu` fork, which is just an alias to `electra` fork

NOTE: create a draft pr right now, do not merge until the spec test is released and passed the tests,


hangleang:
since there is a strong support to move peerdas to fulu for the next devnet, so adding the fork is the first step, and I will make another PR to move the peerdas activation into fulu later

PR #60, opened 2024-12-12
Remove unused consolidation request root from post electra body trait

hangleang:
since there is no used of this trait function, better to remove it from the trait


Tumas:
Hi,

Thanks! I rebased and merged this manually to keep history clean.

PR #67, opened 2024-12-21
EIP7251: Add missed exit checks to consolidation processing

hangleang:
reference spec changes: ethereum/consensus-specs#4000


hangleang:
closing in favor of implementation in 567bc50

PR #69, opened 2024-12-22
Feat: `bls` crate with modular `backend`s

### LINKED ISSUE #61: Add support for `bls12_381` and other crates in our `bls` crate
Currently, we support only `blst` in our `bls` crate. For some use cases we would to have a pure Rust BLS implementation instead of a highly optimized (ASM) BLST. A particular use case is Risc0 and SP1 ZKVMs. SP1 already support BLS precompile via `bls12_381` [patched crate](https://github.com/sp1-patches/bls12_381). Risc0 is planning to implement BLS precompile via the same `bls12_381` patched crate. So my suggestion would be to start with `bls12_381` support and then try to add the rest.

Our [rust-kzg](https://github.com/grandinetech/rust-kzg) we already have multiple ECC backends support. Maybe it even makes sense to extend `rust-kzg` with BLS signatures functionality...

shreyas-londhe:
This PR generalizes the use of backends in the `bls` crate. New backends can be added to this crate and can be used by replacing features where the crate is used.

Resolves #61


shreyas-londhe:
Hi @sauliusgrigaitis I have implemented the necessary traits for the `bls` crate and added the `blst` backend as default. Can you please review my changes and let me know of any changes or suggestions you have? 

I would proceed to add more backends starting with `bls12_381` once you have given a ✅ to my current changes. 

Thank you!

povi:
**[Suggestion](https://conventionalcomments.org/)**:

You don't need to prefix all the trait names with `Bls*`.
I guess the reason for adding the prefixes was that without them trait names clash with other `bls` type names.
You could drop the `Bls*` prefix from trait names and import them like this:

```
use bls::{
    traits::{PublicKey as _, SignatureBytes as _},
    PublicKey, SignatureBytes,
};
```

We rarely need imported trait names in code. Most of the time importing traits anonymously is better.

povi:
**[Suggestion](https://conventionalcomments.org/)**:

All the old code that was wrapping the `blst` crate types was moved to `blst` backend, and the new traits are just the method signatures and nothing more.

When you add a new `bls` backend, much of the code that wraps some newly added bls crate's types will probably be pretty much copy-pasted from `blst` backend with minor changes.

So my suggestion is to try and move as much as possible code from `blst` backend to the trait methods, so it can be reused with different backends.

sauliusgrigaitis:
Thanks @shreyas-londhe for PR, but I agree with @povi. I think it's best if you proceed by adding `bls12_381` crate and then you simply refactor your PR so it doesn't repeat the code.

shreyas-londhe:
Thanks for the suggestions @povi. I'll add the `bls12_281` crate and then move the repeated code to the traits.

shreyas-londhe:
Hi @povi @sauliusgrigaitis, the `bls12_381` crate doesn't have a `hash_to_curve` implementation OOTB, are you guys aware of any existing ones we can use or shall I implement it from scratch?

shreyas-londhe:
Hi @povi I made the changes you requested. Can you please tell me how I can test if my changes didn't break anything? Please let me know of any more suggestions you have :)

sauliusgrigaitis:
@shreyas-londhe the very first step would be to fix the build. You can run the build locally by executing the commands from the build script https://github.com/grandinetech/grandine/blob/develop/.github/workflows/ci.yml.

shreyas-londhe:
@sauliusgrigaitis I'm getting clippy errors which are not from the code I modified. Can you please check that once? I even tried running the tests locally and they fail at code which I have not modified. Can you please help me out on why the CI might be failing?

transition_functions/src/phase0/block_processing.rs
```diff
@@ -485,15 +486,15 @@ mod spec_tests {
...
     // Test files for `process_block_header` are named `block.*` and contain `BeaconBlock`s.
-    processing_tests! {
+    processing_tests!(
         process_block_header,
         |_, state, block: Phase0BeaconBlock<_>, _| unphased::process_block_header(state, &block),
         "block",
         "consensus-spec-tests/tests/mainnet/phase0/operations/block_header/*/*",
-        "consensus-spec-tests/tests/minimal/phase0/operations/block_header/*/*",
-    }
+        "consensus-spec-tests/tests/minimal/phase0/operations/block_header/*/*"
```
povi: I see a TON of changes that are either a space removed, a comma removed or a semicolon added. Neither of these changes are related to this pull request.

What is more, some of these changes you made break macros, like this one. Now it produces error:

```
error: unexpected end of macro invocation
   --> transition_functions/src/phase0/block_processing.rs:494:80
    |
432 |     macro_rules! processing_tests {
    |     ----------------------------- when calling this macro
...
494 |         "consensus-spec-tests/tests/minimal/phase0/operations/block_header/*/*"
    |                                                                                ^ missing tokens in macro arguments
    |
note: while trying to match `,`
   --> transition_functions/src/phase0/block_processing.rs:438:34
    |
438 |             $minimal_glob:literal,
```

povi: Why did you make these changes?

shreyas-londhe: @povi Looks like these were because of running ‘cargo fmt —all’. Let me remove those.

sauliusgrigaitis:
> @sauliusgrigaitis I'm getting clippy errors which are not from the code I modified. Can you please check that once? I even tried running the tests locally and they fail at code which I have not modified. Can you please help me out on why the CI might be failing?

The build on `develop` branch passes https://github.com/grandinetech/grandine/commits/develop/ . If you don't know which of your changes crashed the build -  the easiest way for you would be to apply your changes incrementally and see which change crashed the build.

transition_functions/src/phase0/block_processing.rs
shreyas-londhe: @povi Thanks for pointing this out, I reverted all the formatting changes which are not related to this PR. How do I invoke the CI workflow on my latest commit? I cannot run the CI on my local because I have limited resources and it's taking a lot of time for me to download the submodules.

sauliusgrigaitis: @shreyas-londhe you should be able to fork the repo and push the changes to branches in your local repo. This should trigger CI.

shreyas-londhe:
Hi @povi @sauliusgrigaitis I fixed all the clippy issues and made sure that the CI passes. Can you please review my PR and let me know of any suggestions?

sauliusgrigaitis:
We want to make this inline with https://github.com/grandinetech/rust-kzg naming. So with the same switch both our `bls` and `rust-kzg` will be able to switch to the particular backend. So the name `zkcrypto` needs to be used instead of `bls12_381`. The same is with the other backends such as `arkworks`.

Can you also squash to a single commit after the naming is changed?

sauliusgrigaitis:
I also suggest to to enable `zkcrypto` backend when running tests in Gitlab CI either on Macos or on Windows target.

sauliusgrigaitis:
> I also suggest to to enable `zkcrypto` backend when running tests in Gitlab CI either on Macos or on Windows target.

I just checked and it doesnt compile with `--features bls12_381`. So it really makes sense to add to Gitlab CI. Simply add it [here](https://github.com/grandinetech/grandine/blob/develop/.github/workflows/ci.yml#L37).

shreyas-londhe:
@sauliusgrigaitis The code compiles when `default` features are off. I have kept `bls-blst` as the default backend and other backends can be compiled by running `--no-default-features --features <bls-backend>`. Please let me know if this is desired or I should change it have no default feature.

sauliusgrigaitis:
@shreyas-londhe  you must test your code before you ask for review. It fails https://github.com/grandinetech/grandine/commit/b3577bc0e4fb32f19d5cd008edf87578daf654f1

shreyas-londhe:
Sorry for the confusion @sauliusgrigaitis, as you requested I changed the feature names to match `rust-kzg`.

```bash
# in `grandine` root
cd bls

# builds with `blst` backend
cargo build 

# builds with `zkcrypto (bls12_281)` backend
cargo build --no-default-features --features bls-zkcrypto
```

sauliusgrigaitis:
> Sorry for the confusion @sauliusgrigaitis, as you requested I changed the feature names to match `rust-kzg`.
> 
> ```shell
> # in `grandine` root
> cd bls
> 
> # builds with `blst` backend
> cargo build 
> 
> # builds with `zkcrypto (bls12_281)` backend
> cargo build --no-default-features --features bls-zkcrypto
> ```

The switch needs to work on the top level, not in `bls` crate. You need to edit `.github/workflows/ci.yml` and add something like this https://github.com/shreyas-londhe/grandine/compare/feat/modular-bls...grandinetech%3Agrandine%3Afeat/modular-bls . Then push this change to your fork and make sure that the build with enabled `zkcrypto` passes.

shreyas-londhe:
Thanks for pointing this out @sauliusgrigaitis. The latest commit on my branch passes the CI tests with both `bls-blst` and `bls-zkcrypto` features enabled [here](https://github.com/shreyas-londhe/grandine/blob/feat/modular-bls/.github/workflows/ci.yml#L40). Please let me know if you have any more suggestions. 

I am also planning to write a consistency check macro for testing the newly added backends with the `blst` backend so the issues I faced won't be repeated when more backends are added.

povi:
Looks good! There are some minor things though that I'd suggest to change, I'll write them down.

bls/Cargo.toml
```diff
@@ -6,24 +6,11 @@ authors = ["Grandine <info@grandine.io>"]
...
-zeroize = { workspace = true }
+[features]
+bls-blst = ["dep:bls-blst"]
+bls-zkcrypto = ["dep:bls-zkcrypto"]
 
-[dev-dependencies]
-std_ext = { workspace = true }
-tap = { workspace = true }
+[dependencies]
+bls-core = { path = "bls-core" }
```
povi: **[suggestion](https://conventionalcomments.org/)**: add `bls-core`, `bls-blst` and `bls-zkcrypto` creates next to `bls` crate in top-level `Cargo.toml` file. That way, you only need to specify the paths to those crates only once, and write `= { workspace = true }` everywhere else. No need to use relative paths everywhere.

bls/bls-blst/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+serde_utils = { workspace = true }
+ssz = { workspace = true }
+static_assertions = { workspace = true }
+typenum = { workspace = true }
+zeroize = { workspace = true }
+
+[dev-dependencies]
+std_ext = { workspace = true }
+tap = { workspace = true }
+
```
povi: **[issue](https://conventionalcomments.org/)**: Please remove empty newlines from `*.toml` files that you've added.

bls/bls-core/src/traits/signature.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn verify(&self, message: impl AsRef<[u8]>, public_key: &Self::PublicKey) -> bool;
+
+    #[must_use]
+    fn aggregate(mut self, other: Self) -> Self {
+        self.aggregate_in_place(other);
+        self
+    }
+
+    fn aggregate_in_place(&mut self, other: Self);
+    fn fast_aggregate_verify<'keys>(
```
povi, line 22: **[nitpick](https://conventionalcomments.org/)**: Keep spacing between method signatures consistent. In this case, please add empty lines in between.

bls/bls-core/src/error.rs
```diff
@@ -0,0 +1,23 @@
+use blst::BLST_ERROR;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("invalid public key")]
+    InvalidPublicKey,
+    #[error("invalid secret key")]
+    InvalidSecretKey,
+    #[error("invalid signature")]
+    InvalidSignature,
```
povi, line 8: **[issue](https://conventionalcomments.org/)**: `InvalidSignature` error is not used anywhere. Please remove

bls/bls-blst/src/signature.rs
```diff
@@ -44,9 +43,12 @@ impl TryFrom<SignatureBytes> for Signature {
...
 }
 
-impl Signature {
+impl SignatureTrait for Signature {
+    type SignatureBytes = SignatureBytes;
+    type PublicKey = PublicKey;
+
     #[must_use]
-    pub fn verify(self, message: impl AsRef<[u8]>, public_key: PublicKey) -> bool {
+    fn verify(&self, message: impl AsRef<[u8]>, public_key: &Self::PublicKey) -> bool {
```
povi: **[nitpick](https://conventionalcomments.org/)**: `PublicKey` trait is `Copy`, it does not need to be passed as a reference. Pass `public_key` by value, as it was before.

povi (changes requested):

bls/bls-core/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "bls-core"
+edition = { workspace = true }
+
+[dependencies]
+blst = { workspace = true }
```
sauliusgrigaitis: We need to get rid of `blst` dependency here. Otherwise we still need to compile `blst` if we don't use `blst` as a backend.

bls/Cargo.toml
```diff
@@ -6,24 +6,11 @@ authors = ["Grandine <info@grandine.io>"]
...
-rand = { workspace = true }
-serde = { workspace = true }
-serde_utils = { workspace = true }
-ssz = { workspace = true }
-static_assertions = { workspace = true }
-thiserror = { workspace = true }
-typenum = { workspace = true }
-zeroize = { workspace = true }
+[features]
+bls-blst = ["dep:bls-blst"]
```
sauliusgrigaitis: Is there a particular reason why `bls` prefix is used for feature names? I prefer just `blst`, `zkcrypto`. As I mentioned previously we most likely will try to also hook `rust-kzg` backends using the same switches.

sauliusgrigaitis:
Did you find any way to use `blst` backend by default without explicitly specifying it always? It's very inconvenient.

bls/bls-blst/src/signature.rs
shreyas-londhe: @povi Thank you for the suggestion, but isn't it better if the object is passed by reference where it can be used without the need of copying. I would love to know the thinking behind this.

bls/Cargo.toml
shreyas-londhe: @sauliusgrigaitis I was working with this [suggestion](https://github.com/grandinetech/grandine/pull/69#issuecomment-2567448882), but I realise that you mean the crate names should be prefixed with `bls` and not the feature names. I will change that.

bls/bls-core/src/error.rs
shreyas-londhe: @povi When we remove the `BLST_ERROR` from `bls-core`, `InvalidSignature` is used in `signature.rs`. This change is reflected in the latest commit - https://github.com/shreyas-londhe/grandine/blob/feat/modular-bls/bls/bls-blst/src/signature.rs#L42

bls/bls-blst/src/signature.rs
povi: It depends on the size and the type of the data. Does it make sense to pass a large `struct` by reference? Absolutely. Does it make sense to pass `usize` by reference? Definitely not.

In this case, `PublicKey` is very small, fixed size `struct`. Performance-wise, pushing and popping few times to and from the stack is arguably better than accessing the heap by a reference. In addition to this, code is simpler and cleaner - no need to deal with references.

shreyas-londhe: Thanks for such a detailed explanation, it makes sense. I will make the change.

shreyas-londhe:
> Did you find any way to use blst backend by default without explicitly specifying it always? It's very inconvenient.

@sauliusgrigaitis I'm not sure about this one. I tried using `workspace.features` but they don't seem to link the features correctly. @povi Do you have any suggestions on this one?

povi:
> @povi Do you have any suggestions on this one?

No, sorry, I don't have

shreyas-londhe:
@sauliusgrigaitis I have one solution using `workspace.metadata.commands`, but I don't think that is suitable in this case.

ArtiomTr:
Hey @shreyas-londhe, I've faced the same issue in #88, and solved it by specifying `blst` as default feature in `grandine` crate. Try this patch:

```diff
From 10a946bb54ab25156bdbbc40a31d72933b098dfa Mon Sep 17 00:00:00 2001
From: sirse <artiom.tretjakovas2@gmail.com>
Date: Thu, 23 Jan 2025 19:36:41 +0200
Subject: [PATCH] Mark blst as default feature

---
 grandine/Cargo.toml | 2 ++
 1 file changed, 2 insertions(+)

diff --git a/grandine/Cargo.toml b/grandine/Cargo.toml
index 65f9077..d7387f3 100644
--- a/grandine/Cargo.toml
+++ b/grandine/Cargo.toml
@@ -73,6 +73,8 @@ tempfile = { workspace = true }
 test-case = { workspace = true }
 
 [features]
+default = ["bls/blst"]
+
 logger-always-write-style = []
 
 # `preset-any` and `network-any` should not be passed to Cargo.
-- 
2.43.0
```

shreyas-londhe:
@ArtiomTr Does your patch enable us to switch to different backends at compile time once a default has set?
Something like here: https://github.com/shreyas-londhe/grandine/blob/feat/modular-bls/.github/workflows/ci.yml#L40

ArtiomTr:
Yes, you can do something like:
```
cargo test --release --no-fail-fast --no-default-features --features zkcrypto
```

shreyas-londhe:
@ArtiomTr Thanks for the suggestion. @sauliusgrigaitis @povi I have updated the CI to use the default `bls` backend `blst` and we can explicitly specify the `zkcrypto` backend by calling `--no-default-features --features bls/zkcrypto`. 

I hope this is satisfactory functionality. Please let me know if anything else is required for this PR to be merged.

bls/src/secret_key.rs
```diff
@@ -1,98 +0,0 @@
...
-
-use blst::min_pk::SecretKey as RawSecretKey;
-use derive_more::Debug;
-use serde::Serialize;
-use ssz::{SszHash, SszWrite};
-use static_assertions::assert_not_impl_any;
-
-use crate::{consts::DOMAIN_SEPARATION_TAG, Error, PublicKey, SecretKeyBytes, Signature};
-
-// `RawSecretKey` already implements `Zeroize` (with `zeroize(drop)`):
```
povi, line 16: This pull request is ready to merge except for one last thing: `bls` crate lost most of the comments like this during the refactor. Could you put them back in appropriate places? Thanks

shreyas-londhe: @povi Thanks for pointing this out. I've added the comments back into relevant places.

sauliusgrigaitis (approved):

povi (approved):
Looks good. Thank you for your contribution!

PR #79, opened 2025-01-07
Fetch blobs from EL via `engine_getBlobsV1`:

Tumas:
- Add 'engine_exchangeCapabilities' request to be able to tell if EL serves blobs.
- Fetch capabilities on startup and on every epoch start.
- Rewrite `eth1_api::Endpoints` not to use mutexes.

### DIFF
--- a/eth1_api/src/endpoints.rs
+++ b/eth1_api/src/endpoints.rs
@@ -1,25 +1,20 @@
+use core::sync::atomic::{AtomicBool, Ordering};
+use std::{collections::HashSet, sync::Arc};
+
+use arc_swap::ArcSwap;
 use derive_more::Debug;
+use itertools::Itertools as _;
 use types::redacting_url::RedactingUrl;
 
-#[derive(Clone, Copy, Debug)]
-#[cfg_attr(test, derive(PartialEq, Eq))]
-pub enum EndpointStatus {
-    Online,
-    Offline,
-}
+const ORDERING: Ordering = Ordering::SeqCst;
 
-impl EndpointStatus {
-    const fn is_offline(self) -> bool {
-        matches!(self, Self::Offline)
-    }
-}
-
-#[derive(Clone, Debug)]
-#[cfg_attr(test, derive(PartialEq, Eq))]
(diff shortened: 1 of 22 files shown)


fork_choice_control/src/mutator.rs
```diff
@@ -519,6 +526,11 @@ where
                         Ok(ValidationOutcome::Ignore(false)),
                     );
 
+                    self.request_blobs_from_execution_engine(
+                        pending_block.block.clone_arc(),
+                        missing_blob_indices.clone(),
+                    );
+
                     let blob_ids = missing_blob_indices
```
hangleang, line 534: so in the case if a node missed some blobs, it fetches those blobs from local EL while at the same time, request from peer through RPC as well. why don't implement fallback instead?

Tumas: I believe the idea here is to have blobs available as soon as possible so that block could be processed with minimal delay

eth1_api/src/endpoints.rs
```diff
@@ -1,39 +1,43 @@
...
-    pub const fn is_fallback(&self) -> bool {
-        self.index > 0
+    pub fn is_online(&self) -> bool {
+        self.is_online.load(ORDERING)
+    }
+
+    pub fn set_capabilities(&self, capabilities: HashSet<String>) {
+        if **self.capabilities.load() != capabilities {
+            self.capabilities.store(Arc::new(capabilities));
+        }
```
weekday-grandine-io: [**suggestion (non-blocking)**](https://conventionalcomments.org/): Call `ArcSwap::store` unconditionally.

The load and compare seem redundant.
Is there a reason for comparing the new value with the old one?

```suggestion
        self.capabilities.store(Arc::new(capabilities));
```

Tumas: Capabilities should rarely change, so ArcSwap::store should seldom be needed.  However, this micro-optimization is questionable and it's probably better to avoid extra cognitive load.

povi (approved):
lgtm

PR #82, opened 2025-01-09
Revert BlobSidecarsByRoot/Range version bump

hangleang:
this PR undo #62 based on last ACD call decision, while the `max_blobs_per_block` and `max_request_blob_sidecars` check still there with `electra` values

### DIFF
--- a/p2p/src/network.rs
+++ b/p2p/src/network.rs
@@ -926,8 +926,7 @@ impl<P: Preset> Network<P> {
                 self.handle_blobs_by_range_request(peer_id, peer_request_id, request_id, request)
             }
             RequestType::BlobsByRoot(request) => {
-                self.handle_blobs_by_root_request(peer_id, peer_request_id, request_id, request);
-                Ok(())
+                self.handle_blobs_by_root_request(peer_id, peer_request_id, request_id, request)
             }
             RequestType::Goodbye(goodbye_reason) => {
                 debug!("received GoodBye request (peer_id: {peer_id}, reason: {goodbye_reason:?})");
@@ -980,12 +979,7 @@ impl<P: Preset> Network<P> {
         debug!("received BeaconBlocksByRange request (peer_id: {peer_id}, request: {request:?})");
 
         let start_slot = request.start_slot();
-        let max_request_blocks = match request {
-            OldBlocksByRangeRequest::V1(_) => self.controller.chain_config().max_request_blocks,
-            OldBlocksByRangeRequest::V2(_) => {
-                self.controller.chain_config().max_request_blocks_deneb
-            }
-        };
+        let max_request_blocks = request.max_request_blocks(self.controller.chain_config());
         let difference = request
             .count()
             .min(max_request_blocks)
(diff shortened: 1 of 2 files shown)


eth2_libp2p
Tumas, line 1: Can you make a separate merge request in eth2_libp2p to merge these changes to devnet-5 branch? https://github.com/grandinetech/eth2_libp2p/tree/devnet-5

p2p/src/network.rs
```diff
@@ -1051,28 +1046,21 @@ impl<P: Preset> Network<P> {
...
-            BlobsByRangeRequest::V1(_) => self.controller.chain_config().max_request_blob_sidecars,
-            BlobsByRangeRequest::V2(_) => {
-                self.controller
-                    .chain_config()
-                    .max_request_blob_sidecars_electra
-            }
-        };
-        let difference = request
-            .count()
+        let BlobsByRangeRequest { start_slot, count } = request;
```
Tumas, line 1048: We should use `chain_config().max_request_blob_sidecars` if `chain_config.phase_at_slot::<P>(start_slot + count) < Phase::Electra` (or use  similar check) to correctly validate requested count in Deneb

Tumas (changes requested):

p2p/src/network.rs
Tumas: Ideally, we could implement something similar to [this example](https://github.com/grandinetech/grandine/blob/devnet5/types/src/config.rs#L731), where `max_request_blobs` is statically matched to a specific fork. This way, when a new fork is added, we ensure it is automatically updated, reducing the risk of oversight.

eth2_libp2p
hangleang: I created a PR here: https://github.com/grandinetech/eth2_libp2p/pull/5

p2p/src/network.rs
hangleang: should I apply the same thing to `BlobsByRootRequest` as well?

hangleang: > We should use `chain_config().max_request_blob_sidecars` if `chain_config.phase_at_slot::<P>(start_slot + count) < Phase::Electra` (or use similar check) to correctly validate requested count in Deneb

for check phase, should we check with `start_slot` or `end_slot + 1` as you suggested? if we choose the later, what about the request for a epoch at the fork boundary?

Tumas: Yes, that would be great

hangleang: @Tumas, for `BlobsByRootRequest`, can I use `self.controller.phase()`? since normally, it was requested within the current slot, or at most, current epoch.

Tumas: Yeah, I guess that's probably the best option for `BlobsByRootRequest`.

hangleang: Thanks @Tumas for give it a look, I have updated the `max_request_blob_sidecars` checking with your suggestion

p2p/src/network.rs
```diff
@@ -1127,18 +1116,15 @@ impl<P: Preset> Network<P> {
...
-        let blob_ids = request.blob_ids();
+        let BlobsByRootRequest { blob_ids } = request;
 
         let controller = self.controller.clone_arc();
         let network_to_service_tx = self.network_to_service_tx.clone();
+        let max_request_blob_sidecars = self
+            .controller
+            .chain_config()
+            .max_request_blob_sidecars(self.controller.phase())
+            .expect("cannot request blob_sidecars_by_root before Deneb");
```
Tumas: [**issue**](https://conventionalcomments.org/): in case `expect` is called, it will make app crash by panicking. We use it only for cases where it is extremely unlikely or something is really wrong and application should stop working.
It should not depend on user input (in this case network request)

[**suggestion**](https://conventionalcomments.org/): make `handle_blobs_by_root_request` return a `Result` and return a custom error in this case.

p2p/src/network.rs
```diff
@@ -1051,28 +1046,22 @@ impl<P: Preset> Network<P> {
...
-                    .max_request_blob_sidecars_electra
-            }
-        };
-        let difference = request
-            .count()
+        let BlobsByRangeRequest { start_slot, count } = request;
+        let chain_config = self.controller.chain_config();
+        let max_request_blob_sidecars = chain_config
+            .max_request_blob_sidecars(chain_config.phase_at_slot::<P>(start_slot))
+            .expect("cannot request blob_sidecars_by_range before Deneb");
```
Tumas: [**issue**](https://conventionalcomments.org/): This should not use an `expect`. See my other comment.

p2p/src/network.rs
hangleang: I thought it's supposed to be blocked in the underlying networking layer of `eth2_libp2p` via supported_protocols. I will change accordingly.

Tumas: That's true, but I'd get rid off `expect` in any case so that we or users wouldn't get a surprise in production if there was a bug in `eth2_libp2p`. Error message in the logs is more appropriate.

Tumas (approved):

PR #84, opened 2025-01-10
Optimize hexadecimal conversions in serde_utils to speed up JSON serialization of byte collections

weekday-grandine-io:
See individual commit messages for details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -300,6 +300,7 @@ bytesize = { version = '1', features = ['serde'] }
 cached = '0.53'
 chrono = '0.4'
 clap = { version = '4', features = ['derive'] }
+const-hex = '1.14'
 const_format = '0.2'
 constant_time_eq = '0.3'
 conv = '0.3'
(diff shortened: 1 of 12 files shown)


povi:
lgtm

povi (approved):
lgtm

Cargo.toml
```diff
@@ -339,7 +340,6 @@ hash_hasher = '2'
 hashlink = '0.9'
 hex = { version = '0.4', features = ['serde'] }
```
Tumas, line 341: [**thought**](https://conventionalcomments.org/): `const-hex` claims to be a drop-in replacement for the hex crate, maybe we should drop `hex` and switch to `const-hex` for the rest of the app? Though this might be out of scope for this MR.

Tumas (approved):

Cargo.toml
weekday-grandine-io: You summarized my thought process, including the part about scope.

Another reason[^1] why I'm reluctant to do it is that the richer API of `const-hex` might allow for something better than a drop-in replacement.

[^1]: A bad one.

Issue #59, opened 2024-12-12
Offload states during long non-finalisation


sauliusgrigaitis:
There is an opportunity to reduce memory consumption during long non-finalisation periods by off-loading to the disk first states in the epoch (the rest is already pruned). The current code expects these states to stay in the memory (in chainlinks) so we also need to add a mechanism that is able on demand to load from disk/transit required pruned states that are not in the memory anymore.

PR #87, opened 2025-01-14
Set max amount of epochs for state cache to retain states

povi:
Set max amount of epochs for state cache to retain states, in order to avoid high memory usage during long non-finalization periods.

This is related to #59

### DIFF
--- a/fork_choice_control/src/queries.rs
+++ b/fork_choice_control/src/queries.rs
@@ -6,7 +6,7 @@ use arc_swap::Guard;
 use eth2_libp2p::GossipId;
 use execution_engine::ExecutionEngine;
 use fork_choice_store::{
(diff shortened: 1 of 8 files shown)


sauliusgrigaitis:
This makes deep reorgs complicated. I think it makes sense to benchmark higher amounts of epochs in memory and set a value that makes deeps reorgs easier.

povi:
I've increased the max amount of states to retain in state cache to 64. In non-finality benchmark, this increased the memory usage from 3.6 GiB to 6.4 GiB, which is still reasonable amount of memory used. This should make deep-reorg events easier for client to handle.

povi:
I'm also thinking about adding a CLI argument that overrides this max epochs amount if needed

Tumas (approved):

PR #88, opened 2025-01-15
Added support for multiple KZG backends

### LINKED ISSUE #74: Add support for multiple `rust-kzg` backends
`rust-kzg` supports multiple ECC backends, however, the current approach is by manually switching backends in `Cargo.toml` file. Obviously, it's not convenient. We should have an option to specify at compile time which ECC backends we want to compile in and then allow select one at runtime (via command line flag). Similarly, like #61 for BLS signatures.

It's likely that the best place to do it is https://github.com/grandinetech/rust-kzg/tree/main/kzg . In such case `kzg` crate needs to refactored in order to get rid of `blst` dependency. If we must use some particular ECC backend type (`blst` currently) then we need to use a highly portable ECC backend (pure Rust). Otherwise, if `kzg` crate is not a good place for multibackend support then we should do it in a wrapper in Grandine. 

@ArtiomTr

ArtiomTr:
In this PR:
* Added possibility to optionally compile `blst` or `zkcrypto` backends for kzg_utils package. Both can be included at the same time, and then selected at the runtime.
* Added option to `fork_choice_store::StoreConfig` - `kzg_backend`
* Added CLI option `--kzg-backend`.

Closes #74.

### DIFF
--- a/kzg_utils/Cargo.toml
+++ b/kzg_utils/Cargo.toml
@@ -7,7 +7,11 @@ authors = ["Grandine <info@grandine.io>"]
 [dependencies]
 anyhow = { workspace = true }
 kzg = { workspace = true }
-rust-kzg-blst = { workspace = true }
+rust-kzg-arkworks5 = { workspace = true, optional = true }
+rust-kzg-blst = { workspace = true, optional = true }
+rust-kzg-constantine = { workspace = true, optional = true }
+rust-kzg-mcl = { workspace = true, optional = true }
+rust-kzg-zkcrypto = { workspace = true, optional = true }
 thiserror = { workspace = true }
 types = { workspace = true }
 
@@ -20,3 +24,14 @@ test-generator = { workspace = true }
 
 [lints]
 workspace = true
+
+[features]
+# `bls-backend-any` should not be passed to Cargo.
+# It only exist to avoid duplicating lists of features.
+bls-backend-any = []
+
+arkworks = ["bls-backend-any", "dep:rust-kzg-arkworks5"]
+blst = ["bls-backend-any", "dep:rust-kzg-blst"]
+constantine = ["bls-backend-any", "dep:rust-kzg-constantine"]
+mcl = ["bls-backend-any", "dep:rust-kzg-mcl"]
+zkcrypto = ["bls-backend-any", "dep:rust-kzg-zkcrypto"]
--- a/kzg_utils/src/spec_tests/tests.rs
+++ b/kzg_utils/src/spec_tests/tests.rs
@@ -1,6 +1,5 @@
 #![expect(clippy::manual_let_else)]
 
-use kzg::{eip_4844::bytes_to_blob, Fr as _, G1 as _};
 use spec_test_utils::Case;
 use test_generator::test_resources;
 use types::preset::Mainnet;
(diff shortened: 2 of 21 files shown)


kzg_utils/Cargo.toml
```diff
@@ -20,3 +21,11 @@ test-generator = { workspace = true }
 
 [lints]
 workspace = true
+
+[features]
+# `backend-any` should not be passed to Cargo.
+# It only exist to avoid duplicating lists of features.
+backend-any = []
```
sauliusgrigaitis: I would like something like "bls-backend-any" (I prefer this more) or "ecc-backend-any" (I prefer this less).

sauliusgrigaitis:
The build crashes. Is it the same issue as we have in #69  ? The issue in #69  is that there is no way to make a default backend as `blst`, and the only solution is to provide the backend every time explicitly.

ArtiomTr:
> The build crashes. Is it the same issue as we have in #69 ? The issue in #69 is that there is no way to make a default backend as `blst`, and the only solution is to provide the backend every time explicitly.

Yes, currently build fails as no backends are specified. But I believe this could be fixed by specifying `blst` backend as default feature? I'll try to fix this

sauliusgrigaitis:
> But I believe this could be fixed by specifying blst backend as default feature? I'll try to fix this

Seems there was some issue with setting a default feature. Let us know your findings.

ArtiomTr:
Not exactly sure why it works like that, but default feature needs to be specified in `grandine` crate, not in `kzg_utils`. 

If `blst` feature is defined as default inside `kzg_utils` crate, then running `cargo build --no-default-features` still compiles `kzg_utils` with `blst` feature, and there is no other way of disabling it.

kzg_utils/src/spec_tests/tests.rs
```diff
@@ -32,10 +33,28 @@ fn test_blob_to_kzg_commitment(case: Case) {
     let output = test.output.as_ref().expect("test output should exist");
     let expected_commitment = deserialize(output).expect("should decode commitment bytes");
 
-    let commitment =
-        blob_to_kzg_commitment::<Mainnet>(&blob).expect("should compute blob to kzg commitment");
-
-    assert_eq!(commitment, expected_commitment);
+    let backends = [
```
povi, line 17: **[issue](https://conventionalcomments.org/)**: Trying to run tests with `zkcrypto` backend results in compile errors. Try running:

`cargo test --release --lib -p kzg_utils --features zkcrypto`

results in

```
error[E0433]: failed to resolve: use of undeclared crate or module `rust_kzg_blst`
 --> kzg_utils/src/spec_tests/mod.rs:1:5
  |
1 | use rust_kzg_blst::types::{fr::FsFr, g1::FsG1};
  |     ^^^^^^^^^^^^^ use of undeclared crate or module `rust_kzg_blst`
```

kzg_utils/src/spec_tests/tests.rs
```diff
@@ -58,16 +77,42 @@ fn test_compute_blob_kzg_proof(case: Case) {
...
-        Ok(proof) => {
-            let output = test.output.as_ref().expect("test output should exist");
-            let expected_proof =
-                deserialize(output).expect("should deserialize test output to proof");
-
-            assert_eq!(proof, expected_proof);
-        }
-        Err(_) => {
-            assert!(test.output.is_none());
+    let backends = [
```
povi: **[issue (non-blocking)](https://conventionalcomments.org/)**: This code block is repeated six times in the file. I suggest you move backends list and length check to a separate method. Something like this:

```
fn available_backends() -> impl Iterator<Item = KzgBackend> {
    let backends = [
        #[cfg(feature = "blst")]
        KzgBackend::Blst,
        #[cfg(feature = "zkcrypto")]
        KzgBackend::Zkcrypto,
    ];

    assert_ne!(
        backends.len(),
        0,
        "no backend selected - please provide at least one backend (blst or zkcrypto)"
    );

    backends.into_iter()
}
```

and then call it: `for backend in available_backends() {`

povi (changes requested):

kzg_utils/src/spec_tests/tests.rs
povi: Also, it would be great if build automatically ran tests for both backends

sauliusgrigaitis:
I have a couple of comments:

1) I suggest to add all the backends that `rust-kzg` supports (not only `blst` and `zkcrypto`) including GPU optimised and even `mcl` (it should be merged very soon I think).

2) We now build on Linux, Windows and Macos on Gitlab CI. I suggest to enable different backends for those 3 different builds. But we need to choose which ones. My initial idea would be to use `blst` on Linux, and then choose a couple of other backends for Windows and Macos. `zkcrypto` would be a great candidate for Windows or Macos because we already have zkVM usecase for it. `constantine` or `arkworks` could be the third.

kzg_utils/src/spec_tests/tests.rs
ArtiomTr: I've refactored tests - there was some additional checks before running actual eip_4844 functions, e.g.: https://github.com/grandinetech/grandine/blob/c0678b06a5ab2145db4d5e450747092ed7d62b71/kzg_utils/src/spec_tests/tests.rs#L53-L56

I think these checks were needed previously, as rust-kzg was panicking, when invalid parameters given. But now rust-kzg returns errors, so these checks are no longer required. But if you think that we should keep them, I can bring them back

ArtiomTr: > Also, it would be great if build automatically ran tests for both backends

You mean when running `cargo test` all backends should be tested? Or adding `--features zkcrypto` to CI is enough?

kzg_utils/src/spec_tests/tests.rs
ArtiomTr: Fixed this 👍

sauliusgrigaitis:
@ArtiomTr - can we also add GPU optimised backends?

sauliusgrigaitis:
Seems the build doesn't pass on your branch. Can you rebase and fix the build?

povi (approved):

PR #92, opened 2025-01-20
Prune additional data from storage, add `min_epochs_for_block_requests` to `Config`

Tumas:


fork_choice_control/src/mutator.rs
```diff
@@ -2336,27 +2336,50 @@ where
...
-        );
-        let up_to_slot = misc::compute_start_slot_at_epoch::<P>(up_to_epoch);
+        let blobs_up_to_epoch = self.store.min_checked_data_availability_epoch();
+        let blobs_up_to_slot = misc::compute_start_slot_at_epoch::<P>(blobs_up_to_epoch);
+        let blocks_up_to_epoch = self.store.min_checked_block_availability_epoch();
+        let blocks_up_to_slot = misc::compute_start_slot_at_epoch::<P>(blocks_up_to_epoch);
 
         Builder::new()
-            .name("old-blob-pruner".to_owned())
+            .name("old-record-pruner".to_owned())
```
povi: **[nitpick](https://conventionalcomments.org/)**: Choose a shorter thread name. At least on Linux thread names have 16 byte limit[^1] (including terminating null byte - so that leaves 15 bytes for the name) - anything longer will be truncated.

[^1]: [pthread_setname_np](https://man7.org/linux/man-pages/man3/pthread_getname_np.3.html)

fork_choice_control/src/storage.rs
```diff
@@ -397,6 +397,71 @@ impl<P: Preset> Storage<P> {
         Ok(())
     }
 
+    pub(crate) fn prune_old_blocks_and_states(&self, up_to_slot: Slot) -> Result<()> {
+        let mut block_roots_to_remove = vec![];
+        let mut keys_to_remove = vec![];
+
+        let results = self
+            .database
+            .iterator_descending(..=BlockRootBySlot(up_to_slot).to_string())?;
```
povi: Is this correct? From the first glance, it looks like it will prune data from `up_to_slot` slot, which it shouldn't.

fork_choice_control/src/storage.rs
```diff
@@ -931,10 +1044,65 @@ pub fn serialize(key: impl Display, value: impl SszWrite) -> Result<(String, Vec
...
+        assert_eq!(storage.finalized_block_count()?, 2);
+        assert_eq!(storage.unfinalized_block_count()?, 3);
+        assert_eq!(storage.block_root_by_slot_count()?, 4);
+        assert_eq!(storage.slot_by_state_root_count()?, 2);
+        assert_eq!(storage.state_count()?, 2);
+
+        storage.prune_old_blocks_and_states(5)?;
+
+        assert_eq!(storage.finalized_block_count()?, 0);
+        assert_eq!(storage.unfinalized_block_count()?, 1);
```
povi: I guess it should be 2 here.

fork_choice_control/src/storage.rs
Tumas: good catch!

povi (approved):

PR #94, opened 2025-01-22
Add standard storage mode and improve back-sync handling (30):

### LINKED ISSUE #27: Add a “Goldilocks” prune mode
Keep the required amount and no more, as an alternative to pruning aggressively or not at all.

“Ethereum consensus nodes are required to keep a minimum of 5 months of historical block data ensuring the health of the network.”

I am told that’s `MIN_EPOCHS_FOR_BLOCK_REQUESTS`

Tumas:
Store modes and back-sync:

- Back-sync blocks to `Config::min_epochs_for_block_requests` in standard storage mode
- Back-sync blob sidecars to `Config::min_epochs_for_blob_sidecars_requests`
- Track & filter peers that don't serve blocks prior to `Config::min_epochs_for_block_requests` when performing full back-sync
- Remove `Feature::TrustBackSyncBlocks`
- Verify signatures of back-synced blocks
- Move back-sync status to `Store`
- Relocate `received_blob_sidecars` and `received_block_roots` caches from `p2p::Network` to `p2p::BlockSyncService`
- Extend `SyncBatch` with `retry_count` and `responses_received` fields
- Use smaller back-sync batches when syncing with blobs
- Don't validate signature of genesis block
- Track state archival progress in database to be able to resume it after restart
- Don't request data from busy peers

DB:

- Add `db-info` command to inspect the Sync database
- Replace read-only boolean flag with more descriptive `DatabaseMode` enum

Other:

- Panic to trigger app-restart if network thread is down
- Handle exit signal in an archiver thread & batch archiver updates to db
- Rename `RequestType` to `RPCRequestType` as it conflicts with updated `eth2_libp2p`
- Log peer reporting in debug log
- Log minimal back-sync info when starting back-sync
- Don't log RPC received blocks to info logs (too much output during syncing)

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -20,6 +20,41 @@ use unwrap_none::UnwrapNone as _;
 const GROWTH_STEP: ByteSize = ByteSize::mib(256);
 const MAX_NAMED_DATABASES: usize = 10;
 
+#[derive(Clone, Copy)]
+pub enum DatabaseMode {
+    ReadOnly,
+    ReadWrite,
+}
+
+impl DatabaseMode {
+    #[must_use]
+    pub const fn is_read_only(self) -> bool {
+        matches!(self, Self::ReadOnly)
+    }
+
+    #[must_use]
+    pub const fn mode_permissions(self) -> u16 {
+        match self {
+            // <https://erthink.github.io/libmdbx/group__c__opening.html#gabb7dd3b10dd31639ba252df545e11768>
+            // The UNIX permissions to set on created files. Zero value means to open existing, but do not create.
+            Self::ReadOnly => 0,
+            Self::ReadWrite => 0o600,
+        }
+    }
+
+    #[must_use]
+    #[cfg(target_os = "linux")]
+    pub fn permissions(self) -> u32 {
+        self.mode_permissions().into()
+    }
+
+    #[must_use]
+    #[cfg(not(target_os = "linux"))]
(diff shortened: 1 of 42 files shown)


database/src/lib.rs
```diff
@@ -20,21 +20,56 @@ use unwrap_none::UnwrapNone as _;
...
+    }
+
+    #[must_use]
+    #[cfg(target_os = "linux")]
+    pub fn permissions(self) -> u32 {
+        self.mode_permissions().into()
+    }
+
+    #[must_use]
+    #[cfg(not(target_os = "linux"))]
```
povi, line 52: **[question]**: Is this the same for macOS and windows?

fork_choice_control/src/mutator.rs
```diff
@@ -439,7 +443,10 @@ where
             self.spawn_preprocess_head_state_for_next_slot_task();
         }
 
-        if self.store.is_forward_synced() && misc::slots_since_epoch_start::<P>(tick.slot) == 0 {
+        if self.store.is_forward_synced()
+            && self.store.is_back_synced()
```
povi: **[issue]**: This `if` block contains code that does two things: prune old records and track epoch transition metrics. I think store does not need to be fully back-synced to do the latter. My suggestion is to move `self.store.is_back_synced()` check to the `if tick.kind == TickKind::AttestFourth` clause.

povi (approved):

PR #115, opened 2025-02-17
WIP: Test all KZG backends on all targets in CI

sauliusgrigaitis:


sauliusgrigaitis:
@ArtiomTr the build started to fail when I merged #88. I suggest to try to make all backends passing on all platforms and then disable those that we can't make work.

ArtiomTr:
I think `bls` crate doesn't support both zkcrypto & blst at the same time - only one must be picked now. If we want to maximize checked backends, we can enable these:
```yaml
include:
  - os: ubuntu-latest
    backend: arkworks,blst,constantine,mcl # no zkcrypto (because of blst)
  - os: windows-latest
    backend: arkworks,constantine,zkcrypto # no blst (because of zkcrypto) and mcl
  - os: macos-latest
    backend: arkworks,blst,mcl # no zkcrypto & constantine
```
So on rust-kzg side, `mcl` backend needs fixes for windows, and `constantine` for macos. To make blst & zkcrypto work at the same time, refactoring for `bls` crate is needed.

shreyas-londhe:
Hi @ArtiomTr! Currently the `bls` crate is written such that the backends can be enabled by mentioning the backend as a feature flag. It is by design that we cannot enable multiple backends simultaneously, which imo is the best practice to follow as we do not want any conflicts at runtime because of backend discrepancies. 

Please let me know if you think there can be a better solution to this.

ArtiomTr:
@shreyas-londhe, I think we can make same thing as with `kzg` backends - at compile time, you can choose several backends (e.g. specify `--features zkcrypto,blst`), and then in runtime you can select which backend to choose with command line flag (`grandine --kzg-backend zkcrypto`, maybe it's better to rename it to `--bls-backend` or just `--bls`). In theory, this should work, as all backends work the same way, all have to read same inputs and produce same outputs.

However, one problem may occur, that backends can have different internal representations. Maybe we can abstract this with traits?

PR #123, opened 2025-02-22
Feat: add more `bls` backends

shreyas-londhe:
This PR builds on #69 adding more backends to the `bls` crate. 

Checklist of backends added:
- [x] `arkworks`
- [ ] `constantine`
- [ ] `mcl`

@sauliusgrigaitis Please let me know if some more backends should be added to this list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1068,9 +1068,36 @@ dependencies = [
 name = "bls"
 version = "0.0.0"
 dependencies = [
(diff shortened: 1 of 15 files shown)


sauliusgrigaitis:
@shreyas-londhe I think we should be good with those main 5 - BLST, Arkworks, Zkcrypto, MCL and Constantine.

PR #130, opened 2025-03-13
Enable multiple BLS backends in pipeline

ArtiomTr:
In #88, support for multiple BLS backend in kzg was implemented. However, it broke pipeline, so multiple backend testing was disabled in main branch. This PR enables multiple backends back, without breaking CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,11 +9,11 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
           - os: ubuntu-latest
-            backend: blst
+            backend: arkworks,blst,constantine,mcl
           - os: windows-latest
-            backend: blst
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,18 +9,32 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
           - os: ubuntu-latest
-            backend: blst
+            backend: arkworks,blst,constantine,mcl
           - os: windows-latest
-            backend: blst
+            backend: arkworks,constantine,zkcrypto
```
sauliusgrigaitis, line 14: why no blst on `windows-latest`?

ArtiomTr: because when `zkcrypto` is enabled, it conflicts with `blst` in `bls` crate

PR #136, opened 2025-03-20
Properly stop all services on shutdown

povi:


runtime/src/runtime.rs
```diff
@@ -704,16 +709,32 @@ pub async fn run_after_genesis<P: Preset>(
         result = wait_for_signal() => result,
     }?;
 
+    if stop_clock_tx.send(()).is_err() {
```
Tumas, line 712: Seems like leftovers from testing

povi: It's not. It's the only thing that stops clock service

Tumas (approved):

PR #148, opened 2025-04-02
Attestation pool: reject irrelevant old attestations

povi:


operation_pools/src/attestation_agg_pool/conversion.rs
```diff
@@ -12,17 +13,25 @@ use types::{
...
     InvalidAggregationBits(#[source] ReadError),
+    #[error("attestation is not relevant anymore: {attestation:?}")]
+    Irrelevant { attestation: Box<Attestation<P>> },
 }
 
 pub fn convert_attestation_for_pool<P: Preset, W: Wait>(
     controller: &ApiController<P, W>,
     attestation: Attestation<P>,
 ) -> Result<Phase0Attestation<P>> {
+    if attestation.data().slot + P::SlotsPerEpoch::U64 < controller.head_slot() {
```
sauliusgrigaitis: Shouldn't here we use a wall clock slot?

povi: The idea here was if we are using head state to validate attestations, we should make sure attestations should be validatable using said state. But I agree, clock slot makes more sense here

Tumas (approved):

PR #151, opened 2025-04-02
Allow long state slot processing distances for block validations but only during block sync

povi:
.

### DIFF
--- a/fork_choice_store/src/state_cache_processor.rs
+++ b/fork_choice_store/src/state_cache_processor.rs
@@ -19,6 +19,7 @@ use types::{
 use crate::{Storage, Store};
 
 const ALLOWED_EMPTY_SLOTS_MULTIPLIER: u64 = 2;
+const ALLOWED_EMPTY_SLOTS_MULTIPLIER_FOR_BLOCK_SYNC: u64 = 100;
 
 pub struct StateCacheProcessor<P: Preset> {
     state_cache: StateCache<P>,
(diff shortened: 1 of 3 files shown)


fork_choice_store/src/state_cache_processor.rs
```diff
@@ -86,6 +87,33 @@ impl<P: Preset> StateCacheProcessor<P> {
...
+    // The exception is block sync - which should be allowed, because it's
+    // the only way for the chain to progress in long non-finality events.
+    pub fn try_state_at_slot_for_block_sync<S: Storage<P>>(
+        &self,
+        store: &Store<P, S>,
+        block_root: H256,
+        slot: Slot,
+    ) -> Result<Option<Arc<BeaconState<P>>>> {
+        let allowed_empty_slots_multiplier = if store.is_forward_synced() {
+            ALLOWED_EMPTY_SLOTS_MULTIPLIER
```
sauliusgrigaitis: This does not allow deep reorg. 

I'm not sure that we need this entire `if`. Maybe we whould just use `ALLOWED_EMPTY_SLOTS_MULTIPLIER_FOR_BLOCK_SYNC`.

povi: This opens risks that beacon node during normal operations will receive a block from really outdated node and do thousands of slots of processing.

O the other hand, during chain splits if they are big enough that long slot processing distance is required, beacon node will be out of sync anyway

fork_choice_store/src/state_cache_processor.rs
```diff
@@ -86,6 +87,27 @@ impl<P: Preset> StateCacheProcessor<P> {
...
+            ALLOWED_EMPTY_SLOTS_MULTIPLIER,
+            should_print_slot_processing_warning(store),
+        )
+    }
+
+    // Processing long slot distances are expensive and should not be allowed.
+    // (especially when the processed state is needed for things like attestations
+    // - that could lead to excessive mem and CPU usage and result in DoS).
+    // The exception is block sync - which should be allowed, because it's
+    // the only way for the chain to progress in long non-finality events.
```
sauliusgrigaitis: "in long non-finality events" should be replaced with "long periods without blocks"

sauliusgrigaitis (approved):

PR #152, opened 2025-04-02
Penalize peers that send invalid blocks during block sync

povi:


p2p/src/network.rs
```diff
@@ -327,6 +327,14 @@ impl<P: Preset> Network<P> {
                         P2pMessage::PublishBlobSidecar(blob_sidecar) => {
                             self.publish_blob_sidecar(blob_sidecar);
                         },
+                        P2pMessage::PenalizePeer(peer_id, mutator_rejection_reason) => {
+                            self.report_peer(
+                                peer_id,
+                                PeerAction::LowToleranceError,
```
sauliusgrigaitis, line 333: Does it make sense to do medium or even high tolerance here? The scenario we have in mind is that some nodes on the wrong chain will send wrong blocks, but in this case they will send multiple of those so low tolerance is not necessary.

povi: We can do mid tolerance here

Tumas (approved):

PR #164, opened 2025-04-08
chore: fix `into_blinded` method for proper BeaconBlock handling

lipperhey:
Updated the `into_blinded` method to use `if let` for correctly checking and processing `BeaconBlock`. If not a `BeaconBlock`, it returns the original `self`. Added handling for post-Bellatrix blocks, including `execution_payload`, `kzg_commitments`, and `execution_requests`, before returning a new `blinded_block`.


povi:
In this pull request there is only one change: let ... else { ... } was substituted with if { ... } else { ... } block, both of which are correct and do the same. This pull request changes nothing.

> Updated the into_blinded method to use if let for correctly checking and processing BeaconBlock.

It was correct before.

>  Added handling for post-Bellatrix blocks, including execution_payload, kzg_commitments, and execution_requests

You added nothing.

PR #220, opened 2025-06-03
Fix missing blobs sync after fulu activated

hangleang:
this PR include:
* fix missing blobs sync after fulu activated
* reconstruct using data column sidecars from cached and storage
* request data columns by root to the most coverage peer if any
* ~~request data columns by range to a subset of available peers, leaving some for by root requests~~

### DIFF
--- a/fork_choice_control/src/controller.rs
+++ b/fork_choice_control/src/controller.rs
@@ -564,13 +564,10 @@ where
         let store_snapshot = self.store_snapshot();
         if let Some(block_root) = store_snapshot.get_delayed_block_at_slot(slot) {
             if !store_snapshot.is_sidecars_construction_started(block_root) {
-                let available_columns_count =
-                    store_snapshot.available_columns_at_block(*block_root).len();
+                let accepted = store_snapshot.accepted_data_column_sidecars_at_slot(slot);
 
-                if available_columns_count < store_snapshot.sampling_columns_count()
-                    && available_columns_count > 0
-                    && available_columns_count * 2
-                        >= store_snapshot.chain_config().number_of_columns()
+                if accepted < store_snapshot.sampling_columns_count()
+                    && accepted * 2 >= store_snapshot.chain_config().number_of_columns()
                 {
                     MutatorMessage::ReconstructMissingColumns {
                         wait_group: self.owned_wait_group(),
@@ -751,6 +748,10 @@ where
 
     pub fn sampling_columns(&self) -> HashSet<ColumnIndex> {
         self.store_snapshot().sampling_columns().clone()
+    }
+
+    pub fn sampling_columns_count(&self) -> usize {
+        self.store_snapshot().sampling_columns().len()
     }
 
     pub fn accepted_data_column_sidecar(
--- a/p2p/src/sync_manager.rs
+++ b/p2p/src/sync_manager.rs
@@ -27,7 +27,11 @@ use typenum::Unsigned as _;
 use types::{
     config::Config,
     deneb::containers::BlobIdentifier,
-    fulu::{consts::NumberOfColumns, containers::DataColumnIdentifier, primitives::ColumnIndex},
+    fulu::{
+        consts::NumberOfColumns,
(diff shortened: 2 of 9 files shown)


fork_choice_control/src/controller.rs
```diff
@@ -564,13 +564,11 @@ where
...
-                let available_columns_count =
-                    store_snapshot.available_columns_at_block(*block_root).len();
+                let accepted = store_snapshot.accepted_data_column_sidecars_at_slot(slot);
 
-                if available_columns_count < store_snapshot.sampling_columns_count()
-                    && available_columns_count > 0
-                    && available_columns_count * 2
-                        >= store_snapshot.chain_config().number_of_columns()
+                if accepted < store_snapshot.sampling_columns_count()
+                    && accepted > 0
```
povi: `accepted > 0` line is not needed. The line below already checks that there is sufficient amount of accepted data columns

fork_choice_control/src/controller.rs
```diff
@@ -564,13 +564,11 @@ where
...
+                    && accepted * 2 >= store_snapshot.chain_config().number_of_columns()
```
povi, line 571: I'm trying, but failing to understand this line: doesn't `store_snapshot.chain_config().number_of_columns()` define how many data sidecar columns there is in total for any given block? So how can there be more than double of the total of possible data columns accepted at a slot?

fork_choice_control/src/controller.rs
hangleang: updated with the comment

povi (approved):

p2p/src/sync_manager.rs
```diff
@@ -592,35 +592,37 @@ impl SyncManager {
...
+                    }
 
                     if !columns_to_request.is_empty() {
+                        // TODO(peerdas-fulu): Distributes requested columns among the minimal set of peers,
+                        // selected peers should be checked prior the assignment to make sure those peers
+                        // can serve all of the requested columns.
+                        let end_peer_index = (start_peer_index
+                            + columns_to_request.len().min(MAX_REQUEST_PEERS_PER_EPOCH))
+                        .min(peers_to_sync.len());
+                        let peers_to_request = &peers_to_sync[start_peer_index..end_peer_index];
```
hangleang: Hey @povi, I still hesitate to merge this PR because I feel like it should be more dynamic, and selected peers SHOULD cover at least most of the request columns by checking their advertised custody columns, rather than slice array like this. Do you have any idea or suggestion?

povi: So we have a set of peers connected, and a set of columns to request, and we can assign one peer per column index, ideally covering as much column indices as possible. This is [maximum bipartite matching](https://www.geeksforgeeks.org/maximum-bipartite-matching/) problem.

If you want as close to optimal solution as possible, what you can do here is try something like [`hopcroft-karp`](https://crates.io/crates/hopcroft-karp) crate (or some other crate that solves this). (Though I don't like that this crate does not have much downloads)

On the other hand, maybe some more naive, simplier algorithm would be enough here?

hangleang: > So we have a set of peers connected, and a set of columns to request, and we can assign one peer per column index, ideally covering as much column indices as possible. 

currently, we do that. but that is not efficient for utilize peers that way, then we have no available (non-busy) for further requests (by root requests for example) 

> If you want as close to optimal solution as possible, what you can do here is try something like [`hopcroft-karp`](https://crates.io/crates/hopcroft-karp) crate (or some other crate that solves this).
> 
> On the other hand, maybe some more naive, simplier algorithm would be enough here?

I would like to have a peers columns mapping algorithms, which can use to load-balance the request load among its peers (not all into a single peer, not all peers in an epoch requests)

hangleang: I have been exclude the slice array from this pull request, as there is no significant improvement with the change. we can address this at later pull request, avoid blocking other fixes from inclusion

PR #221, opened 2025-06-04
Implement basic validator custody without backfill

hangleang:
In this PR: 

* Update peer columns custody after CGC in peer metadata changed

* Update custody requirements at the start of epoch

* Trigger update custody group count in ENR and metadata based on updates schedule

Limitation: delayed advertising current custody group count till next epoch after restart

### DIFF
--- a/eip_7594/src/lib.rs
+++ b/eip_7594/src/lib.rs
@@ -24,9 +24,9 @@ use types::{
         containers::{DataColumnSidecar, MatrixEntry},
         primitives::{CellsAndKzgProofs, ColumnIndex, CustodyIndex},
     },
-    phase0::primitives::{NodeId, SubnetId},
+    phase0::primitives::{Gwei, NodeId, SubnetId, ValidatorIndex},
     preset::Preset,
-    traits::SignedBeaconBlock as _,
+    traits::{BeaconState, SignedBeaconBlock as _},
 };
 
 use error::Error;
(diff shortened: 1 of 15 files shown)


eip_7594/src/lib.rs
```diff
@@ -134,6 +134,32 @@ pub fn compute_subnets_for_node(
...
 }
 
+#[must_use]
+pub fn compute_custody_subnets_and_columns_for_node(
+    raw_node_id: [u8; 32],
+    custody_group_count: u64,
+    config: &Config,
+) -> (HashSet<SubnetId>, HashSet<ColumnIndex>) {
+    let custody_groups = get_custody_groups(raw_node_id, custody_group_count, config)
+        .expect("should compute custody groups for node");
```
povi: Please use `expect` only in cases when you're 100% sure the code cannot crash (unless it's in tests). And write the reason why you think it is safe to unwrap and it won't panic. "should compute custody groups for node" does not explain why we should `expect` that the code won't crash.

Suggested change:
```
#[must_use]
pub fn compute_custody_subnets_and_columns_for_node(
    raw_node_id: [u8; 32],
    custody_group_count: u64,
    config: &Config,
) -> Result<(HashSet<SubnetId>, HashSet<ColumnIndex>)> {
    let custody_groups = get_custody_groups(raw_node_id, custody_group_count, config)?;
    ...
```

hangleang: Updated with the suggestion!

Ignore `eth2_libp2p` conflict for now until other PR merged

validator/src/custody.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let write_to_disk = || -> Result<()> {
+        let filename = dir.join(CUSTODY_UPDATES_SCHEDULE_FILENAME);
+
+        let mut file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(filename)?;
+
+        writeln!(file, "{epoch},{custody_group_count}")?;
```
povi, line 82: All good, except I really don't like storing data in plain files. I'd suggest you create a new database using `Database` from local `database` crate. It is used across the app, you can find real code examples how to use it.

The reasons for using database instead of plain files:
* We have `persistent` and `in_memory` modes. It would be great when app runs in `in_memory` mode files are not created.
* Easier portability between different OS.
* It handles (internally) various failure cases and concurrent accesses.

But since this is already done, refactoring this may be out of scope of this pull request. I'll approve this pull request anyway, it's up to you to choose to do the refactor (files -> DB) now, or add a `TODO` and do it later.

povi (approved):

validator/src/custody.rs
hangleang: Thank you for suggestions, 

I will choose to do it later, and not merging it now. Since it is still an optional for `devnet-1`, the spec discussion is still going on, and the implementation still has the limitation mentioned in the description.

PR #230, opened 2025-06-12
Implement Status V2 with with earliest_available_slot

povi:


hangleang (approved):
~~All good! but do we need unit test for the `StatusV2`~~ nvm I saw that being tested in `eth2_libp2p`

PR #243, opened 2025-07-01
Add `PubkeyCache` that stores decompressed pubkey bytes:

Tumas:
- remove `CachedPublicKey` and its internal cache
- move `PrefixableKey` trait to `database` crate


pubkey_cache/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+
+            for unpersisted in self.unpersisted.iter() {
+                debug!("pubkey {:?} is still unpersisted", *unpersisted);
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn get_or_insert(&self, public_key_bytes: &PublicKeyBytes) -> Result<Arc<PublicKey>> {
```
povi: **[nitpick]**: I'd use `PublicKeyBytes` instead of a reference (it does implement `Copy`)

pubkey_cache/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            database.put_batch(batch)?;
+
+            debug!("stored {entries} validator pubkeys to pubkey cache db");
+        }
+
+        Ok(())
+    }
+
+    pub fn persist<P: Preset>(&self, state: &BeaconState<P>) -> Result<()> {
+        if let Some(database) = self.database.as_ref() {
```
povi: **[nitpick]**: if `if` takes whole method, maybe change it to guard statement:

```
let Some(database) = self.database.as_ref() else {
    return Ok(());
}
```

pubkey_cache/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            let mut batch = vec![];
+
+            for validator in state.validators() {
+                if let Some(pubkey) = self.unpersisted.remove(&validator.pubkey) {
```
povi: Add comment why `self.unpersisted` is checked against `state.validators()` instead of just persisting all unpersisted keys

pubkey_cache/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            if let Some(pubkey) = self.unpersisted.remove(&validator.pubkey) {
+                batch.push(serialize(&PublicKeyDbKey(pubkey), &decompressed));
+            }
+        }
+
+        if let Some(database) = self.database.as_ref() {
+            let entries = batch.len();
+            database.put_batch(batch)?;
+
+            debug!("stored {entries} validator pubkeys to pubkey cache db");
```
povi: **[nitpick]**: change it to `persisted {entries} validator pubkeys to pubkey cache db` as `persist` word is used in this module in all the places.

povi (approved):

PR #247, opened 2025-07-07
Update consensus-spec-tests to v1.6.0-alpha.3

hangleang:


hangleang:
Hi @povi, could you give me an idea why spec tests still failed on `not_available` cases?

I don't get why the `engine_getBlobsV2` request doesn't get triggered in the tests when any of columns missing. what am I missing here?

hangleang:
the new spec tests include [this change](https://github.com/ethereum/consensus-specs/pull/4421/files#diff-d5862f8f60447deb160988defa243b6d74d7a51146d9807f576ae5f760493f50) in the fork choice's `on_block` test for fulu fork

povi:
> Hi @povi, could you give me an idea why spec tests still failed on `not_available` cases?
> 
> I don't get why the `engine_getBlobsV2` request doesn't get triggered in the tests when any of columns missing. what am I missing here?

Checked failing tests, `Store` `sampling_columns` is empty during block validation, so it assumes that block is fully data-available.

hangleang:
> Checked failing tests, `Store` `sampling_columns` is empty during block validation, so it assumes that block is fully data-available.

Ahh! I see, that's why no `engine_getBlobsV2` got called. Thank you

fork_choice_control/src/tasks.rs
```diff
@@ -554,20 +554,14 @@ impl<P: Preset, W> Run for ReconstructDataColumnSidecarsTask<P, W> {
...
+        let available_columns = (0..store_snapshot.chain_config().number_of_columns)
+            .filter_map(|index| {
                 let identifier = DataColumnIdentifier { block_root, index };
-                match store_snapshot.cached_data_column_sidecar_by_id(identifier) {
-                    Some(data_column_sidecar) => Ok(Some(data_column_sidecar)),
-                    None => storage.data_column_sidecar_by_id(identifier),
-                }
+                store_snapshot
+                    .cached_data_column_sidecar_by_id(identifier)
+                    .or_else(|| storage.data_column_sidecar_by_id(identifier).ok().flatten())
```
povi: I'm not a fan of silent failures. At least log a `WARN` on failure.

hangleang: Thanks, I haven't look the details, I thought it will throw an error if any sidecar doesn't exist in storage. rolled it back

povi (approved):

PR #249, opened 2025-07-08
feat: add log_level as cli

barnabasbusa:
* Added LogLevel enum in grandine_args.rs with variants: Error, Warn, Info, Debug, Trace
* Added --log-level CLI flag to the GrandineArgs struct with Info as the default


sauliusgrigaitis:
This approach doesn't work in real world. The volume of logs will be huge. That's why we manually enable particular modules.

barnabasbusa:
Can't be worse than some other client's debug logs

barnabasbusa:
still better than not being able to change the log level without compiling imo.

sauliusgrigaitis:
> still better than not being able to change the log level without compiling imo.

You can set particular module debug level without compiling with a env variable, for example `GRANDINE_LOG=p2p=debug`. That's what we actually use and that's why the logging code is as it is now.

barnabasbusa:
Any reason why we can't have both?

sauliusgrigaitis:
The reason we haven't got it so far is because we never needed ourselves such an option to get other level of logging for _all_ the crates.

PR #255, opened 2025-07-14
Batch data column by root requests

hangleang:
Instead of requesting data columns by root for each block, this pull request will change to batch those by root requests and request to respective custodial peers at once after by range requests finished. doing so will require to track the received data column sidecars for every slot/block root

### DIFF
--- a/p2p/src/block_sync_service.rs
+++ b/p2p/src/block_sync_service.rs
@@ -33,7 +33,11 @@ use tokio_stream::wrappers::IntervalStream;
 use types::{
     config::Config,
     deneb::containers::BlobIdentifier,
-    fulu::containers::{DataColumnIdentifier, DataColumnsByRootIdentifier},
+    fulu::{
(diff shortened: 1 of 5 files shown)


p2p/src/block_sync_service.rs
```diff
@@ -1111,6 +1131,104 @@ impl<P: Preset> BlockSyncService<P> {
...
     }
 
+    fn batch_request_missing_data_columns(&mut self) -> Result<()> {
+        let mut missing_column_by_indices: HashMap<ColumnIndex, HashSet<H256>> = HashMap::new();
+        for ColumnIndicesByRoot {
+            block_root,
+            columns,
+        } in self.sync_manager.get_missing_columns_by_root()
+        {
+            // let  = column_indices_by_root;
```
povi: Remove development artifacts

hangleang: removed!

povi (approved):

PR #272, opened 2025-07-24
Add more PeerDAS metrics

hangleang:


prometheus_metrics/src/metrics.rs
```diff
@@ -63,13 +63,15 @@ pub struct Metrics {
...
     pub data_column_sidecar_verification_times: Histogram,
-    pub reconstructed_columns: IntCounter, // TODO
+    pub reconstructed_columns: IntCounter,
     pub columns_reconstruction_time: Histogram,
     pub data_column_sidecar_computation: Histogram,
     pub data_column_sidecar_inclusion_proof_verification: Histogram,
-    pub data_column_sidecar_kzg_verification_single: Histogram, // TODO?
     pub data_column_sidecar_kzg_verification_batch: Histogram,
-    pub beacon_custody_columns_count_total: IntCounter, // TODO
+    pub beacon_custody_columns_count_total: IntCounter,
```
mininny, line 71: It seems this metric isn't implemented yet. I think it would be good to also change this name to `beacon_custody_groups` when it's implemented!

mininny (approved):
I ran this branch in kurtosis with other CLs and checked that the metrics are rendering same as other implementations :+1:

prometheus_metrics/src/metrics.rs
hangleang: yes, agreed

hangleang: @mininny do you know what is difference between `beacon_custody_groups` and `beacon_custody_groups_backfilled` metrics? my sense is custody groups changed when there is a certain increased in EB of attached vaildators, and it get immediately advertise. I see no point to track the backfill here for custody groups

PR #274, opened 2025-07-25
Backfill custody group columns

povi:


hangleang:
Hello @povi, I don't see any uses of peer's `earliest_available_slot` in the backfill process. My understanding is client need to take peer's `earliest_available_slot` into account when requesting columns from that peer, because of its `cgc` get updated immediately when the total effective balance of attached validators increased by a certain number, so it's likely peer doesn't has what they immediately advertised.

povi:
> Hello @povi, I don't see any uses of peer's `earliest_available_slot` in the backfill process. My understanding is client need to take peer's `earliest_available_slot` into account when requesting columns from that peer, because of its `cgc` get updated immediately when the total effective balance of attached validators increased by a certain number, so it's likely peer doesn't has what they immediately advertised.

Resolved

p2p/src/sync_manager.rs
```diff
@@ -303,12 +311,27 @@ impl<P: Preset> SyncManager<P> {
             };
 
             let start_slot = current_back_sync_slot.saturating_sub(count);
+            let count = current_back_sync_slot.saturating_sub(start_slot);
```
hangleang, line 314: all good! except a small nit on missing the head slot (current slot) columns backfill, more details in DM

povi: fixed those issues

Tumas (approved):

hangleang (approved):

PR #294, opened 2025-08-11
Update consensus-spec-tests to v1.6.0-alpha.4

hangleang:
Update spec tests to https://github.com/ethereum/consensus-specs/releases/tag/v1.6.0-alpha.4, there are 2 new SSZ types has been introduced in EIP7916, since those types are not relevant to Fulu, and PeerDAS specifically, I have just excluded from the script test. In this PR, I also (`mostly`) move `NumberOfColumns` into preset to compliant with the [spec change](https://github.com/ethereum/consensus-specs/pull/4485) while keep it as constant in some place where involve too deep changes.

### DIFF
--- a/eip_7594/src/lib.rs
+++ b/eip_7594/src/lib.rs
@@ -145,10 +145,7 @@ pub fn compute_subnets_for_node(
 }
 
 /// Verify if the data column sidecar is valid.
-pub fn verify_data_column_sidecar<P: Preset>(
-    data_column_sidecar: &DataColumnSidecar<P>,
-    config: &Config,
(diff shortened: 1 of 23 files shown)


eth2_libp2p
povi, line 1: Did you push `eth2_libp2p` changes?

hangleang: Ahh! A moment

hangleang: https://github.com/grandinetech/eth2_libp2p/pull/12

povi (approved):
LGTM

PR #297, opened 2025-08-12
Peerdas fulu

### LINKED ISSUE #337: Merge fulu to develop

Tumas:


eth1_api/src/eth1_api.rs
```diff
@@ -167,7 +171,39 @@ impl Eth1Api {
         Ok(None)
     }
 
-    pub(crate) async fn get_blobs<P: Preset>(
+    // pub(crate) async fn get_blobs<P: Preset>(
```
povi: Remove if not needed

binary_utils/src/lib.rs
```diff
@@ -20,8 +20,8 @@ pub fn initialize_logger(module_path: &str, always_write_style: bool) -> Result<
         .filter_module("dedicated_executor", LevelFilter::Info)
         .filter_module("doppelganger_protection", LevelFilter::Info)
         .filter_module("eth1", LevelFilter::Info)
-        .filter_module("eth1_api", LevelFilter::Info)
-        .filter_module("eth2_libp2p", LevelFilter::Info)
+        .filter_module("eth1_api", LevelFilter::Debug)
+        .filter_module("eth2_libp2p", LevelFilter::Debug)
```
povi: Do not log these at debug level by default

grandine/src/main.rs
```diff
@@ -405,6 +405,8 @@ fn try_main() -> Result<()> {
         kzg_backend,
         blacklisted_blocks,
         report_validator_performance,
+        withhold_data_columns_publishing,
+        disable_engine_getblobs,
     } = config;
 
     features.into_iter().for_each(Feature::enable);
```
Tumas: Enabling features here is no longer necessary as they are now enabled earlier in grandine/src/grandine_args.rs:1294

eip_7594/src/error.rs
```diff
@@ -0,0 +1,22 @@
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error(
+        "Custody group count is invalid: {custody_group_count} expected <= {number_of_custody_groups}",
```
Tumas: We have a convention of starting errors with a lowercase letter, unless it's an acronym or a title. We should stick to it. There are other violations of this in other places as well.

binary_utils/src/lib.rs
hangleang: done with recent rebase

eth1_api/src/eth1_api.rs
hangleang: addressed in https://github.com/grandinetech/grandine/pull/297/commits/369d09e91f21c96721d0898c196ec0e56168ff58

grandine/src/main.rs
hangleang: yes, you are right. fix is at https://github.com/grandinetech/grandine/pull/297/commits/369d09e91f21c96721d0898c196ec0e56168ff58

eip_7594/src/error.rs
hangleang: okay, I updated to follow the convention at https://github.com/grandinetech/grandine/pull/297/commits/369d09e91f21c96721d0898c196ec0e56168ff58

validator/src/validator.rs
```diff
@@ -882,6 +898,23 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
...
+                            &signed_blinded_block,
+                        )
+                        .await
+                    {
+                        warn!("failed to post blinded block to the builder node: {error:?}");
+                    }
+
+                    debug!("submitted blinded block to the builder node");
+
+                    return Ok(());
```
Tumas: This returns from the function no matter if there was an error or not when submitting signed blinded block to builder using v2. It might be better to retry with V1 in case V2 is not supported by the builder. 

Also, returning here won't call this code: 

```
        info!(
            "validator {} proposing beacon block with root {:?} in slot {}",
            proposer_index,
            beacon_block.message().hash_tree_root(),
            slot_head.slot(),
        );
```

and 

            
```
        if let Some(metrics) = self.metrics.as_ref() {
            metrics.validator_propose_successes.inc();
        }

```
        
Even though V2 does not return the block and blobs, we should still execute these calls using the blinded block’s root.

I’d also suggest including information about using the builder in the log message, as the code path differs from the existing one.

validator/src/validator.rs
```diff
@@ -941,17 +974,57 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
...
+                if self
+                    .chain_config
+                    .phase_at_slot::<P>(slot_head.slot())
+                    .is_peerdas_activated()
+                {
+                    let cells_and_kzg_proofs = eip_7594::try_convert_to_cells_and_kzg_proofs::<P>(
+                        blobs.as_ref(),
+                        block_proofs.unwrap_or_default().as_ref(),
+                        self.controller.store_config().kzg_backend,
+                    )?;
```
Tumas: missing an empty line

validator/src/validator.rs
hangleang: > This returns from the function no matter if there was an error or not when submitting signed blinded block to builder using v2. It might be better to retry with V1 in case V2 is not supported by the builder.
> 
Yes, there was a discussion whether to implement the fallback when V2 return error https://discord.com/channels/595666850260713488/1395080509775937636/1397717728436949053 . I'm fine if we decide to do that, and I think it is much better to handle such the case on our own, rather than expect builder to do the job.

> Also, returning here won't call this code:
> 
> ```
>         info!(
>             "validator {} proposing beacon block with root {:?} in slot {}",
>             proposer_index,
>             beacon_block.message().hash_tree_root(),
>             slot_head.slot(),
>         );
> ```
> 
> and
> 
> ```
>         if let Some(metrics) = self.metrics.as_ref() {
>             metrics.validator_propose_successes.inc();
>         }
> ```
> 
> Even though V2 does not return the block and blobs, we should still execute these calls using the blinded block’s root.
> 
> I’d also suggest including information about using the builder in the log message, as the code path differs from the existing one.

what kind of additional information do we need to display here? could you give some examples?

Tumas: just "using builder" is enough: `"validator {} proposing beacon block with root {:?} in slot {} using builder"`

block_producer/src/block_producer.rs
```diff
@@ -916,6 +923,80 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
                     },
                 })
             }
+            Phase::Fulu => {
+                // Store results in a vec to preserve insertion order and thus the results of the packing algorithm
```
Tumas: The attestation-building logic for block inclusion is the same in both Electra and Fulu, but is currently duplicated. Given its non-trivial nature, it should be refactored into a shared method for reuse

hangleang: sure, I gonna bundle this with the fallback implement

validator/src/validator.rs
hangleang: please have a look at https://github.com/grandinetech/grandine/pull/300

eth1_api/src/execution_blob_fetcher.rs
```diff
@@ -159,4 +211,181 @@ impl<P: Preset, W: Wait> ExecutionBlobFetcher<P, W> {
...
+        block_or_sidecar: BlockOrDataColumnSidecar<P>,
+        data_column_identifiers: Vec<DataColumnIdentifier>,
+    ) {
+        let slot = block_or_sidecar.slot();
+        let block_root = block_or_sidecar.block_root();
+
+        if self.controller.contains_block(block_root)
+            || self.sidecars_construction_started.contains_key(&block_root)
+        {
+            debug!("cannot fetch blobs from EL: block has been imported, or being importing");
```
Tumas: grammar typo, change to: "cannot fetch blobs from EL: block has been imported, or is being imported"

http_api/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+use dashmap as _;
```
Tumas: Move `DashMap` dependency to `[dev-dependencies]` section in `http_api/Cargo.toml` instead.

types/src/fulu/consts.rs
```diff
@@ -0,0 +1,4 @@
+use typenum::U128;
+
+// TODO(feature/fulu): put it in preset
+pub type NumberOfColumns = U128;
```
Tumas: currently this duplicates `P::NumberOfColumns`. Unless there is a good to reason to keep it, it should be removed

hangleang: it is being used in `types/src/config.rs::columns_per_group` function, which is then being used deep in the stack. migrate to use the preset value would require significant change, but let's me try

hangleang: addressed in https://github.com/grandinetech/grandine/pull/308

http_api/src/standard.rs
```diff
@@ -1189,12 +1204,39 @@ pub async fn blob_sidecars<P: Preset, W: Wait>(
         })
         .collect::<Result<Vec<_>>>()?;
 
-    let blob_sidecars = controller.blob_sidecars_by_ids(blob_identifiers)?;
-    let blob_sidecars = DynamicList::try_from_iter_with_maximum(
-        blob_sidecars.into_iter(),
-        usize::try_from(max_blobs_per_block).map_err(AnyhowError::new)?,
-    )
-    .map_err(AnyhowError::new)?;
+    let blob_sidecars = if version.is_peerdas_activated() {
```
Tumas, line 1242: Not sure how relevant this endpoint is post-Fulu, but the current behavior looks incorrect. It should collect all available column identifiers, not just the first half, and then attempt reconstruction if possible.

It might be useful to provide a method to fetch all available columns by block root.

hangleang: yes, there is a way to reconstruct if we have any more than 64 columns; then return the requested blobs, but first if we have to check if it is supernode, so it should have the first half columns, which we then can construct or align those columns into blobs, so no reconstruction needed. 

also with any 64 columns, is it worth to do reconstruction just to return the blobs every time that endpoint called?

hangleang: > It might be useful to provide a method to fetch all available columns by block root.

we have `/eth/v1/debug/beacon/data_column_sidecars/{block_id}` endpoint for that

Tumas: >  is it worth to do reconstruction just to return the blobs every time that endpoint called?

We should cover both cases (alignment and reconstruction) if endpoint is called post-Fulu.

Tumas: > we have /eth/v1/debug/beacon/data_column_sidecars/{block_id} endpoint for that

I mean it might be useful to encapsulate this logic into a method, as it could be then reused in both this endpoint and the case above.

fork_choice_store/src/store.rs
```diff
@@ -3366,6 +3693,37 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
             .collect()
     }
 
+    pub fn indices_of_missing_data_columns(
+        &self,
+        block: &SignedBeaconBlock<P>,
+    ) -> Vec<ColumnIndex> {
+        let block = block.message();
+
+        let Some(body) = block.body().post_deneb() else {
```
Tumas: should probably be `post_fulu`

http_api/src/standard.rs
hangleang: @Tumas please have a look at https://github.com/grandinetech/grandine/pull/314

fork_choice_store/src/store.rs
```diff
@@ -2047,6 +2318,12 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         }
 
         self.blob_cache.on_slot(new_tick.slot);
+        // TODO(peerdas-fulu): NEED REVIEW!
```
hangleang: need review! assess what could be wrong when it in a situation stated below

fork_choice_store/src/store.rs
```diff
@@ -3366,6 +3693,37 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+            return vec![];
+        };
+
+        if body.blob_kzg_commitments().is_empty() {
+            return vec![];
+        }
+
+        let block_root = block.hash_tree_root();
+
+        // TODO(peerdas-fulu): figure out the way to check sampling columns without storing the indices in sync_manager
```
hangleang: I tried to find a way to reference `sampling_columns` from `network_globals` here, but it introduce cyclic-dependent parameters when both modules (`Network` and `Controller`) got initialized in `runtime/src/runtime.rs`

p2p/src/network.rs
```diff
@@ -158,6 +167,17 @@ impl<P: Preset> Network<P> {
         let (shutdown_tx, shutdown_rx) = futures::channel::mpsc::channel(1);
         let executor = TaskExecutor::new(logger.clone(), shutdown_tx);
 
+        // TODO: get custody value from metadata
```
hangleang: we _should_ get cgc value of metadata from persisted source when the node get restarted, but it is not urgent now

p2p/src/network.rs
```diff
@@ -837,17 +932,43 @@ impl<P: Preset> Network<P> {
...
+            .expect("should compute node custody groups");
+
+        let mut sampling_columns = HashSet::new();
+        for custody_index in &custody_groups {
+            let columns = compute_columns_for_custody_group::<P>(*custody_index, config)
+                .expect("should compute custody columns for node");
+            sampling_columns.extend(columns);
+        }
+
+        // TODO(peerdas-fulu): use `sampling_columns` from `network_globals` in
```
hangleang: if we can reference `sampling_columns` from `network_globals`, we are no longer need to sync the value

fork_choice_control/src/mutator.rs
```diff
@@ -577,63 +630,169 @@ where
...
-                                    pending_blob,
-                                    Some(state.clone_arc()),
+                            if self
+                                .store
+                                .indices_of_missing_data_columns(&pending_block.block)
+                                .is_empty()
+                            {
+                                self.retry_block(wait_group, pending_block);
+                            } else {
+                                // TODO(peerdas-fulu): NEED REVIEW! if block proposed by itself, therefore all sampling
```
hangleang, line 708: need review! is it a proper way to check if the block proposed by itself, or is there any accurate way to do so?

fork_choice_control/src/mutator.rs
```diff
@@ -1904,6 +2359,54 @@ where
...
+        if should_retry_block {
+            if let Some(pending_block) = self.take_delayed_until_blobs(block_root) {
+                self.retry_block(wait_group.clone(), pending_block);
+            }
+        }
+
+        self.event_channels
+            .send_data_column_sidecar_event(block_root, data_column_sidecar);
+
+        // TODO(peerdas-fulu): We can't keep data column sidecars on cache for too long, and we
```
hangleang: link to my comment https://github.com/grandinetech/grandine/pull/297/files#r2300568658

p2p/src/sync_manager.rs
```diff
@@ -434,27 +549,100 @@ impl SyncManager {
...
+                        self.log(
+                            Level::Debug,
+                            format_args!(
+                                "requesting columns ({}): [{}] for slots: {start_slot}..={max_slot}",
+                                missing_column_indices.len(),
+                                missing_column_indices.iter().join(", "),
+                            ),
+                        );
+
+                        // TODO(peerdas-fulu): Distributes requested columns among the minimal set of peers,
```
hangleang: a task for sync optimization which the mapping should be able to assign only a subset of custodial peers to request per epoch, and unassigned peers will be requested for the next epoch, and so on. the current behaviour is map to broader peers, sometime one column per peer which is too optimistic.

The problem with the approach is it will introduce complexity into the codebase

p2p/src/sync_manager.rs
```diff
@@ -434,27 +549,100 @@ impl SyncManager {
...
-            sync_batches.push(SyncBatch {
-                target: SyncTarget::Block,
-                direction: SyncDirection::Forward,
-                peer_id,
-                start_slot,
-                count,
-                response_received: false,
-                retry_count: 0,
-            });
+                // TODO(peerdas-fulu): review this max requests cap
```
hangleang: please review the soft cap for sync request (half of number of peers to sync)

eip_7594/src/lib.rs
```diff
@@ -27,65 +78,360 @@ pub fn get_custody_groups(node_id: NodeId, custody_group_count: u64) -> Vec<Cust
...
+ * Recover the full, flattened sequence of matrix entries.
+ *
+ * This helper demonstrates how to apply ``recover_cells_and_kzg_proofs``.
+ */
+pub fn recover_matrix<P: Preset>(
+    partial_matrix: &[MatrixEntry<P>],
+    blob_count: usize,
+    backend: KzgBackend,
+) -> Result<Vec<MatrixEntry<P>>> {
+    // TODO(peerdas-fulu): group once by row_index
```
hangleang: refactor at https://github.com/grandinetech/grandine/pull/322/commits/352d912eb4829b39a336f6fe0ae6fc5b4baf5f0b

p2p/src/sync_manager.rs
```diff
@@ -434,27 +549,100 @@ impl SyncManager {
...
+                                    "could not find available peers to request data column sidecars".to_owned(),
+                                );
+
+                                self.refresh_custodial_peers();
+                            }
+                        }
+                    }
+                }
+
+                // TODO(feature/eip7594): refactor SyncBatch to Enum instead of struct with options
```
hangleang: should we refactor or leave as it now?

fork_choice_store/src/store.rs
```diff
@@ -2642,6 +2956,19 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
             .retain(|(slot, _, _), _| finalized_slot <= *slot);
+        self.accepted_data_column_sidecars
+            .retain(|(slot, _, _), _| finalized_slot <= *slot);
+        self.sidecars_construction_started
+            .retain(|_, slot| finalized_slot <= *slot);
+        self.delayed_block_at_slot
+            .retain(|slot, _| finalized_slot <= *slot);
+        self.requested_blobs_from_el
+            .retain(|_, slot| finalized_slot <= *slot);
+        // TODO(feature/eip-7594): NEED REVIEW!
```
hangleang: double check on that is this still a concern here?

hangleang (commented):
@Tumas @povi please help review the comments, and suggest what should be done on those TODOs

fork_choice_store/src/store.rs
povi: Spawn persisting task for each data column. My tests showed no issue in doing so.

p2p/src/sync_manager.rs
povi: Optimization would be nice, but right now it's not the top priority. Leave the `TODO` for the future fix

p2p/src/sync_manager.rs
povi: This is not a blocking issue, leave it with `TODO`

fork_choice_store/src/store.rs
povi: If it calls persist task for each data column, then this is not an issue

fork_choice_control/src/mutator.rs
povi: Checking if block origin is `BlockOrigin::Own` is good way to determine if the block is proposed by the same beacon node. If the block was proposed by a local EL, then it has all the sidecars.

However, we still have the issue if block was built using external block builder. In that case beacon node does not have all the needed sidecars and should start the reconstruction.

povi: One way to would to modify `Own` origin to include data if it was built with external builder or not:

```
pub enum BlockOrigin {
    ...
    Own { from_external_builder: bool },
    ...
}
```

or something like that. And then do the reconstruction (or not) depending on the `bool`

hangleang: if the block is built by an external builder node, the proposer node will received the block over the network, right? also, in that case, it won't spawn own `BlockTask`: https://github.com/grandinetech/grandine/blob/5596b39718b559ae80c802c72a9b08549b5c260a/validator/src/validator.rs#L1058. do I get it right?

p2p/src/network.rs
hangleang: done in https://github.com/grandinetech/grandine/pull/344

fork_choice_store/src/store.rs
hangleang: in https://github.com/grandinetech/grandine/pull/344

povi (approved):

PR #301, opened 2025-08-15
fix: add maximum size validation for JWT secrets to prevent DoS attacks

crStiv:
Add maximum size validation for JWT secrets in eth1_api auth module

- Add JWT_SECRET_SIZE_MAX_BYTES constant (1024 bytes) to prevent DoS attacks
- Update size validation to check both minimum (32 bytes) and maximum (1024 bytes) 
- Improve error message to indicate valid size range
- Add test case for large JWT secret rejection

### DIFF
--- a/eth1_api/src/auth.rs
+++ b/eth1_api/src/auth.rs
@@ -20,8 +20,7 @@ use zeroize::Zeroizing;
 #[cfg(test)]
 use derive_more::Debug;
 
-const JWT_SECRET_SIZE_MIN_BYTES: usize = 32;
-const JWT_SECRET_SIZE_MAX_BYTES: usize = 1024; // Reasonable maximum size to prevent DoS attacks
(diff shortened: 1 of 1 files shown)


Tumas:
Hey,

Thanks for the pull request! A few notes:

- The secret is configured and read from the local filesystem, not exchanged over the network, so a DoS attack does not apply in this context.
- [According to the current spec,](https://github.com/ethereum/execution-apis/blob/b050556ffb7d69d9281af8d2ffd4096192715b7c/src/engine/authentication.md) the secret key is expected to be 256 bits. If it is not, clients **should** return an error. [Previously](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.8/src/engine/authentication.md), the requirement was “at least 256 bits,” which explains the current implementation.

Feel free to adjust your pull request or open a new one to enforce an exact 256-bit check for the secret if you prefer.

crStiv:
@Tumas made some changes, hope I implemented everything exactly as you wished it to be done

crStiv:
@Tumas pinging

PR #310, opened 2025-08-20
Fix peerdas-related metrics

hangleang:


http_api/src/standard.rs
```diff
@@ -1262,17 +1262,18 @@ pub async fn publish_block<P: Preset, W: Wait>(
...
             controller.store_config().kzg_backend,
         )?;
 
-        let data_column_sidecars = eip_7594::construct_data_column_sidecars(
-            &signed_beacon_block,
-            &cells_and_kzg_proofs,
-            metrics.as_ref(),
-        )?;
+        let data_column_sidecars =
+            eip_7594::construct_data_column_sidecars(&signed_beacon_block, &cells_and_kzg_proofs)?;
```
Tumas, line 1276: I think timer should be stopped here as it otherwise it will also track publishing time

eth1_api/src/execution_blob_fetcher.rs
```diff
@@ -295,14 +301,15 @@ impl<P: Preset, W: Wait> ExecutionBlobFetcher<P, W> {
...
-                                                self.metrics.as_ref(),
                                             ),
                                             BlockOrDataColumnSidecar::Sidecar(sidecar) => eip_7594::construct_data_column_sidecars_from_sidecar(
                                                 &sidecar,
                                                 &cells_and_kzg_proofs,
-                                                self.metrics.as_ref(),
                                             ),
                                         };
+
+                                        prometheus_metrics::stop_and_record(timer);
```
Tumas, line 312: I see that in other places `data_column_sidecar_computation` tracks combined time of both `eip_7594::try_convert_to_cells_and_kzg_proofs` and `eip_7594::construct_data_column_sidecars` but here it only tracks `eip_7594::try_convert_to_cells_and_kzg_proofs`. Is this intended?

http_api/src/standard.rs
```diff
@@ -1342,16 +1343,18 @@ pub async fn publish_blinded_block<P: Preset, W: Wait>(
...
             controller.store_config().kzg_backend,
         )?;
-        let data_column_sidecars = eip_7594::construct_data_column_sidecars(
-            &signed_beacon_block,
-            &cells_and_kzg_proofs,
-            metrics.as_ref(),
-        )?;
+
+        let data_column_sidecars =
+            eip_7594::construct_data_column_sidecars(&signed_beacon_block, &cells_and_kzg_proofs)?;
```
Tumas, line 1359: I think timer should be stopped here as it otherwise it will also track publishing time

http_api/src/standard.rs
```diff
@@ -1444,16 +1447,18 @@ pub async fn publish_block_v2<P: Preset, W: Wait>(
...
```
Tumas, line 1465: I think timer should be stopped here as it otherwise it will also track publishing time

http_api/src/standard.rs
hangleang: I first thought the same

hangleang: I update based on your comments, have a look @Tumas

eth1_api/src/execution_blob_fetcher.rs
```diff
@@ -287,9 +292,6 @@ impl<P: Preset, W: Wait> ExecutionBlobFetcher<P, W> {
                                     .into_iter()
                                     .flat_map(IntoIterator::into_iter)
                                     .collect::<Vec<_>>();
-                                let timer = self.metrics.as_ref().map(|metrics| {
-                                    metrics.data_column_sidecar_computation.start_timer()
-                                });
```
Tumas: can you revert this part? My original comment was incorrect. Your initial change is correct

hangleang: why?

Tumas: so that timer would start just before `try_convert_to_cells_and_kzg_proofs` as you initially added.

hangleang: okay, now is before the `try_convert_to_cells_and_kzg_proofs` called

Tumas (approved):

PR #314, opened 2025-08-21
HTTP API: reconstruct for blob sidecars if more than half columns are available

hangleang:


http_api/src/standard.rs
```diff
@@ -3889,6 +3881,89 @@ async fn wait_for_missing_blocks_with_timeout<P: Preset, W: Wait>(
...
+        return Ok(vec![]);
+    };
+
+    if data_column_sidecars.len() * 2 < P::NumberOfColumns::USIZE {
+        return Ok(vec![]);
+    }
+
+    let half_columns = P::NumberOfColumns::U64.saturating_div(2);
+
+    if (0..half_columns).any(|index| {
```
Tumas, line 3900: can you briefly explain why this check is necessary

http_api/src/standard.rs
```diff
@@ -3889,6 +3881,89 @@ async fn wait_for_missing_blocks_with_timeout<P: Preset, W: Wait>(
...
+            .map(ByteVector::from)
+            .map(Blob::<P>::from)
+            .map_err(Into::into)
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    misc::construct_blob_sidecars(
+        block,
+        blobs,
+        vec![KzgProof::zero(); body.blob_kzg_commitments().len()],
```
Tumas, line 3963: shouldn't this be a proper proof? instead of `KzgProof::zero`?

hangleang: what do you mean by proper proof?

Tumas: non-empty one

http_api/src/standard.rs
hangleang: oop! I thought of checking any missing columns in the first half. I will update it

http_api/src/standard.rs
Tumas: or is empty proof a valid in this case?

hangleang: yes, the original proposal is to use empty proof. but it has been closed in favor for a new separate API endpoint https://github.com/ethereum/beacon-APIs/pull/524#issuecomment-2892253605

Tumas (approved):

PR #322, opened 2025-08-26
Clean up codebase and temp fix CI pipeline

hangleang:


p2p/src/sync_manager.rs
```diff
@@ -346,7 +346,6 @@ impl<P: Preset> SyncManager<P> {
                                                 count,
                                                 response_received: false,
                                                 retry_count: 0,
-                                                // TODO(feature/fulu): handle error
                                                 data_columns: ContiguousList::try_from(columns)
                                                     .map(Arc::new)
```
Tumas: You could use [`inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) to output info to error log to avoid silent fails.

p2p/src/sync_manager.rs
```diff
@@ -582,7 +581,6 @@ impl<P: Preset> SyncManager<P> {
                                         count,
                                         response_received: false,
                                         retry_count: 0,
-                                        // TODO(feature/fulu): handle error case
```
Tumas, line 593: You could use [`inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) to output info to error log to avoid silent fails.

p2p/src/block_sync_service.rs
```diff
@@ -757,7 +757,6 @@ impl<P: Preset> BlockSyncService<P> {
                                     );
 
                                     for (peer_id, columns) in peer_custody_columns_mapping {
-                                        // TODO(feature/fulu): catch error here
```
povi, line 759: Please give more sufficient reason in `expect` why you're sure `ContiguousList::try_from(columns.clone())` will always 100% succeed (because you're removing TODO; and `expect` will crash block sync service if it fails).

hangleang: should we change the type of `data_columns` to vector instead of `ContiguousList`? I see no real reason for keep using the type internally, and parsing it over and over again. we can change to vector and only parse into `ContiguousList` in `request_batches`

Tumas:
build passes btw: https://github.com/grandinetech/grandine/actions/runs/17238742596

p2p/src/block_sync_service.rs
hangleang: @povi what should I do next to proceed this PR?

povi: > should we change the type of data_columns to vector instead of ContiguousList?

Yes, that's a good idea

povi: However, let's do this later with a new PR

Tumas (approved):

povi (approved):

PR #339, opened 2025-09-02
Update consensus-spec-tests to v1.6.0-alpha.6 (#335)

Tumas:


hangleang (approved):

hangleang:
> `Config::ttfb_timeout` and `Config::resp_timeout` are not removed as they are still used in `eth2_libp2p`

there is a commit for this changes in LH: https://github.com/sigp/lighthouse/commit/7b5be8b1e74cf0ce71aeda71e017dd426920e955

PR #334, opened 2025-09-02
Add `--sync-with-reconstruction` option for greedy sync

hangleang:

PR #348, opened 2025-09-04
Count accepted data column sidecars by block header

### LINKED ISSUE #347: Count `accepted_data_column_sidecars` by `block_header`

hangleang:
Merge this pull request after https://github.com/grandinetech/grandine/pull/334, because it involve changes in prior pull request

### DIFF
--- a/fork_choice_store/src/store.rs
+++ b/fork_choice_store/src/store.rs
@@ -2605,10 +2605,14 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         self.data_column_cache.insert(data_sidecar);
     }
 
-    pub fn accepted_data_column_sidecars_at_slot(&self, slot: Slot) -> usize {
+    pub fn accepted_data_column_sidecars_count(&self, block_header: BeaconBlockHeader) -> usize {
         self.accepted_data_column_sidecars
-            .keys()
-            .filter(|(s, _, _)| *s == slot)
+            .iter()
+            .filter(|((slot, proposer_index, _), commitments)| {
+                *slot == block_header.slot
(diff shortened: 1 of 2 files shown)


fork_choice_store/src/store.rs
```diff
@@ -2605,10 +2605,12 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         self.data_column_cache.insert(data_sidecar);
     }
 
-    pub fn accepted_data_column_sidecars_at_slot(&self, slot: Slot) -> usize {
+    pub fn accepted_data_column_sidecars_count(&self, block_header: BeaconBlockHeader) -> usize {
         self.accepted_data_column_sidecars
             .keys()
-            .filter(|(s, _, _)| *s == slot)
+            .filter(|(slot, proposer_index, _)| {
+                *slot == block_header.slot && *proposer_index == block_header.proposer_index
```
Tumas: this should probably include block_root check as well

hangleang: so, checking `slot` along with `proposer_index` is not accurate enough?

hangleang: checking `block_root` would be challenging as it store in `HashMap`

hangleang: > this should probably include block_root check as well

updated with the `block_root`

Tumas: > so, checking slot along with proposer_index is not accurate enough?

Not sure but in an edge case, where the are two different data columns for different blocks at the same slot signed by the same validator this might incorrectly return the result.

povi (approved):

PR #350, opened 2025-09-04
Add `Store::is_reconstruction_enabled`

Tumas:
Fixes too many block retries issue when syncing without reconstruction

### DIFF
--- a/fork_choice_control/src/mutator.rs
+++ b/fork_choice_control/src/mutator.rs
@@ -2407,15 +2407,10 @@ where
 
         let accepted_data_columns = self.store.accepted_data_column_sidecars_count(block_header);
 
-        // There is no data columns by each root request, while syncing we batch by root requests
-        // to respective custodial peers in `p2p/src/block_sync_service.rs::batch_request_missing_data_columns` method
-        let should_retry_block = if self.store.is_sidecars_construction_started(&block_root)
-            || (!self.store.is_forward_synced()
-                && self.store.sampling_columns_count() * 2 < P::NumberOfColumns::USIZE)
-        {
-            accepted_data_columns == self.store.sampling_columns_count()
-        } else {
+        let should_retry_block = if self.store.is_reconstruction_enabled_for(block_root) {
             accepted_data_columns * 2 >= self.store.sampling_columns_count()
+        } else {
+            accepted_data_columns == self.store.sampling_columns_count()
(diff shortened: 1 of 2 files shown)


fork_choice_control/src/mutator.rs
```diff
@@ -2407,15 +2407,10 @@ where
...
-        // There is no data columns by each root request, while syncing we batch by root requests
-        // to respective custodial peers in `p2p/src/block_sync_service.rs::batch_request_missing_data_columns` method
-        let should_retry_block = if self.store.is_sidecars_construction_started(&block_root)
-            || (!self.store.is_forward_synced()
-                && self.store.sampling_columns_count() * 2 < P::NumberOfColumns::USIZE)
-        {
-            accepted_data_columns == self.store.sampling_columns_count()
-        } else {
+        let should_retry_block = if self.store.is_reconstruction_enabled_for(block_root) {
             accepted_data_columns * 2 >= self.store.sampling_columns_count()
```
hangleang, line 2411: should we make it explicit here "only retry when half of total columns are available, in this case 64", or we keep it dynamic based on the sampling size?

Tumas: I would say to it keep it dynamic which is more generic and would not need any changes if we decided to allow users to customize custody group count

hangleang: yes, with validator custody. the custody group count will increase based on attached effective balances, I'm not sure if there is a plan to allow set the value explicitly

hangleang (approved):

fork_choice_control/src/mutator.rs
Tumas: Oh, I misunderstood the question. Actually "only retry when half of total columns are available, in this case 64" is better here.  E.g. if for whatever reason sampling size is >= 64 and < 128, this would unnecessary retry blocks.

hangleang: but in case of synced, it would need to request from peers for the missing columns, so let say we set it to 64 to retry, while available columns only 63, so it won't retry, though it won't request from peers as well. so, I think keeping it dynamic here wouldn't cause any problems though

Tumas: but if it failed to receive the missing data column it would get stuck

PR #356, opened 2025-09-08
Use `u64::checked_div` when calculating `max_slot_head`

Tumas:
Avoids potential panic


p2p/src/sync_manager.rs
```diff
@@ -1086,37 +1086,39 @@ impl<P: Preset> SyncManager<P> {
...
-            })
-            .into_iter()
-            .filter_map(|(block_root, indices)| {
-                (indices.len() != sampling_count).then(|| {
-                    let missing = controller
-                        .sampling_columns()
-                        .difference(&indices)
-                        .copied()
-                        .collect();
+        Some(
```
povi: On such large code block I would prefer `.pipe(Some)` at the end instead of wrapping whole block in `Some` - less indentation, less changes

povi (approved):

PR #362, opened 2025-09-10
Use separate dedicated executor for data column reconstruction

Tumas:


fork_choice_control/src/mutator.rs
```diff
@@ -710,11 +710,10 @@ where
                                 if !matches!(pending_block.origin, BlockOrigin::Own)
                                     && !self.store.is_sidecars_construction_started(&block_root)
                                 {
-                                    self.handle_reconstructing_data_column_sidecars(
-                                        wait_group,
```
povi, line 714: Double check that tests and `ad_hoc_bench` is not broken due to not waiting

runtime/src/runtime.rs
```diff
@@ -180,6 +181,9 @@ pub async fn run_after_genesis<P: Preset>(
         metrics.clone(),
     ));
 
+    let dedicated_executor_for_reconstruction =
+        DedicatedExecutor::new("de-reconstruction", 1, None, metrics.clone());
```
povi: "de-reconstruction" length is 17 bytes, while linux max thread name length is 15 bytes + null

operation_pools/src/blob_reconstruction_pool/tasks.rs
```diff
@@ -0,0 +1,78 @@
...
+        } = self;
+
+        let Ok(available_columns) = controller.data_column_sidecars_by_ids(
+            (0..P::NumberOfColumns::U64).map(|index| DataColumnIdentifier { block_root, index }),
+        ) else {
+            warn!("failed to retrieve data column sidecar from storage");
+            return Ok(());
+        };
+
+        if available_columns.len() == controller.sampling_columns_count() {
```
povi: TL;DR: Available columns are not necessarily the same columns as node sampling columns. Just checking the count is not sufficient.

Beacon node accepts all the data columns that receives from the network, not just the sampled ones. As we seen before, some other nodes may send even columns with indices beacon node did not request. Since the `available_columns` collects all the columns for same block root (`(0..P::NumberOfColumns::U64).map(|index| DataColumnIdentifier { block_root, index })`), `available_columns` very well may include unsolicited columns too.

povi (approved):

PR #364, opened 2025-09-10
Move EL & beacon API reconstruction code into `tokio::spawn::blocking`

Tumas:
- Affects EL blob fetcher and beacon API

### DIFF
--- a/eth1_api/src/execution_blob_fetcher.rs
+++ b/eth1_api/src/execution_blob_fetcher.rs
@@ -16,6 +16,7 @@ use helper_functions::misc;
 use log::{debug, warn};
 use prometheus_metrics::Metrics;
 use ssz::{ContiguousList, H256};
+use std_ext::ArcExt as _;
 use types::{
     combined::SignedBeaconBlock,
     deneb::{containers::BlobIdentifier, primitives::BlobIndex},
@@ -212,7 +213,6 @@ impl<P: Preset, W: Wait> ExecutionBlobFetcher<P, W> {
         }
     }
 
-    #[expect(clippy::cognitive_complexity)]
     #[expect(clippy::too_many_lines)]
     async fn get_blobs_v2(
         &self,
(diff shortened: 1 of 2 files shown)


eth1_api/src/execution_blob_fetcher.rs
```diff
@@ -229,73 +229,86 @@ impl<P: Preset, W: Wait> ExecutionBlobFetcher<P, W> {
...
+                            let controller = self.controller.clone_arc();
 
-                                let timer = self.metrics.as_ref().map(|metrics| {
-                                    metrics.data_column_sidecar_computation.start_timer()
-                                });
+                            let sidecars_construction_started =
+                                self.sidecars_construction_started.clone_arc();
+
+                            let received_data_column_sidecars =
+                                self.received_data_column_sidecars.clone_arc();
```
hangleang, line 303: will this be able to use to mutate the underlying data through these `Arc<T>` in dedicated thread? sorry, this might be a noop question, I'm not fully understand about synchronous programming, so I just want to clarify on that. it would be helpful if you can link to specific docs about this

Tumas: yes, there will be no problem for mutation as [DashMap::insert](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html#method.insert) takes [shared reference](https://doc.rust-lang.org/reference/interior-mutability.html)

Tumas: also the reason for this change is that long-running non-yielding cpu-heavy tasks executed in regular tokio tasks can block the executor and prevent it from driving other futures, so in general these should be offloaded to a dedicated [thread](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html).

hangleang (approved):

povi (approved):

PR #365, opened 2025-09-11
Fix incorrect fork exists check in fork context

hangleang:


povi (approved):

fork_choice_control/src/mutator.rs
```diff
@@ -2410,7 +2410,7 @@ where
         let should_retry_block = if self.store.is_reconstruction_enabled_for(block_root) {
             accepted_data_columns * 2 >= P::NumberOfColumns::USIZE
         } else {
-            accepted_data_columns == self.store.sampling_columns_count()
+            accepted_data_columns * 2 >= self.store.sampling_columns_count()
```
povi: I don't think this is correct

hangleang: I have ignore this change, only update `eth2_libp2p` to fix bug

PR #371, opened 2025-09-12
Fx data columns by root rpc request limit

hangleang:


eth2_libp2p
Tumas, line 1: merged `eth2_libp2p` PR, please point submodule to `77ee727d9d5f2bdbf464146c672bb484e8b120f2`

hangleang: done

Tumas: can you squash all these changes into 1 commit? It would be easier on history and to revert if ever needed.

Tumas (approved):

PR #380, opened 2025-09-18
Add Gloas boilerplate

hangleang:


block_producer/src/block_producer.rs
```diff
@@ -886,7 +893,8 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
 
                 let attestations = ContiguousList::try_from_iter(attestations)?;
 
-                if self.beacon_state.phase() == Phase::Electra {
+                let phase = self.beacon_state.phase();
+                if phase == Phase::Electra {
```
povi: It's time to use `match` on phase here (instead of several `if-else`)

hangleang: fixed in https://github.com/grandinetech/grandine/pull/380/commits/97afd8ac7c8ae72714649d5ea418bf532a73db7c

hangleang: @povi can you give a thumb up if this PR is fine?

hangleang:
close in favor of #392

PR #386, opened 2025-09-25
Implemented zkvm-pico host and guest programs

jimmychu0807:
In addition to the zkvm-pico host and guest code, there are a few notable issues worth to be reviewed and discussed in this PR, summarized  below.

- **zkvm/host/src/backend.rs** has been refactored to **backend** dir, and each zkvm host program is named as **backend/<zkvm-name>.rs**. Refer to [backend/mod.rs](https://github.com/grandinetech/grandine/pull/386/files#diff-3482dbe68138f25e4b11e40e81b9ec249058b08afc54da0780663ab14f378e79).
- Added [**zkvm/host/src/pico.rs**](https://github.com/grandinetech/grandine/pull/386/files#diff-940f069e4f5c73958e246678533e9a1bc81c972ef970f0eb16b33b49c27d62cb) as pico host program.
- Added [**zkvm/host/pico_helper**](https://github.com/grandinetech/grandine/pull/386/files#diff-478b88b7e4ee1a965391136ab8e534a9fe88058633c8ee07c6b56f2af295b7a1) so compiling the zkvm-host pico program (`cargo build -p zkvm_host --features pico`) will also compile the zkvm-guest pico program.
- In **bls-zkcrypto** package code, added a new [`zkvm-pico` feature](https://github.com/grandinetech/grandine/pull/386/files#diff-b7052b91afd19d705c1fc1a239061689620d6a33549eb6f367fa450e21785eb2) to support running pico library that take a different parameter type on `hash_to_curve()` function in [secret_key.rs](https://github.com/grandinetech/grandine/pull/386/files#diff-bbad1fc544ca0bfa7f5dc97fac2339e2f3f233cb456210f00143e0f3244251af) and [signature.rs](https://github.com/grandinetech/grandine/pull/386/files#diff-d1beff702398e41d25d4f018be88937b50c7246e957523cc7d3e7a5139096f57).
- The [rust-toolchain.toml](https://github.com/grandinetech/grandine/pull/386/files#diff-2b1bde2cf3a858b7bf7424cb8bcbf01f35b94dc80b925d9432cbab3319ca9b4e) has changed to using `nightly-2025-08-04` nightly version. This is necessary to compile pico-sdk for the host and guest program. Need to discuss further.
- The root [**Cargo.toml**](https://github.com/grandinetech/grandine/pull/386/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542), I have commented out universal-precompile for bls12_381 and sha2. Once this PR looks good. I can make another PR of the corresponding bls12_381 and sha2 lib into **grandinetech/universal-precompile**.
- Added a fix on [ssz/src/shared.rs](https://github.com/grandinetech/grandine/pull/386/files#diff-ec44286ad4cbc12cd03c34bb4cc367606de56aaf990c0c2e6b65d3bfe84f1c28) on the actual > maximum check on **read_list()**.
  cc @povi author of the ssz-reading optimization code - please check if the fix is actually proper 🙏, @Dyslex7c
- Added a [.editorconfig](https://github.com/grandinetech/grandine/pull/386/files#diff-0947e2727d6bad8cd0ac4122f5314bb5b04e337393075bc4b5ef143b17fcbd5b) file for better editor display support.

This PR updates rust-toolchain from stable to nightly version. If this is too drastic to the main branch, I can comment out the nightly version line and add a comment that compiling zkvm-pico requires switching from stable rust toolchain to nightly.

@ArtiomTr, @povi please have a look on this PR when you have time. Thanks.


Cargo.toml
```diff
@@ -299,7 +305,10 @@ base64 = '0.22'
 bincode = '1'
 bit_field = '0.10'
 bitvec = '1'
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+# bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+# For Brevis Pico
+bls12_381 = { git = "https://github.com/jimmychu0807/brevis-bls12_381.git", branch = "jc/dev" }
```
ArtiomTr: This change breaks all other zkvms, so need to be sorted out before merging into develop

Cargo.toml
```diff
@@ -433,7 +442,10 @@ serde_with = '3'
 serde_yaml = '0.9'
 # The `asm` feature in `sha2` doesn't do anything on recent `x86_64` CPUs
 # because `sha2` defaults to using CPU intrinsics.
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0', features = ['compress'] }
+# sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0', features = ['compress'] }
+# For Brevis Pico
+sha2 = { git = 'https://github.com/brevis-network/hashes.git', branch = 'sha2-0.10.8-v1.0.0', features = ['compress'] }
```
ArtiomTr: Same as bls crate

bls/bls-zkcrypto/Cargo.toml
```diff
@@ -24,3 +24,6 @@ zeroize = { workspace = true }
 
 [target.'cfg(target_os = "zkvm")'.dependencies]
 rand_chacha = { workspace = true }
+
+[features]
+zkvm-pico = []
```
ArtiomTr: Better not to have special features for zkvms, but to detect them with target, e.g. for risc0:
```rs
#[cfg(target_os = "zkvm", target_vendor = "risc0")]
```

bls/bls-zkcrypto/src/secret_key.rs
```diff
@@ -66,10 +66,17 @@ impl SecretKeyTrait<SIZE> for SecretKey {
...
         let h = <G2Projective as HashToCurve<ExpandMsgXmd<Sha256>>>::hash_to_curve(
             [message.as_ref()],
             DOMAIN_SEPARATION_TAG,
         );
+        #[cfg(feature = "zkvm-pico")]
+        let h = <G2Projective as HashToCurve<ExpandMsgXmd<Sha256>>>::hash_to_curve(
+            message.as_ref(),
+            DOMAIN_SEPARATION_TAG,
+        );
+
```
ArtiomTr: I believe that it is old zkcrypto API, that was in last published version `0.8.0`, but grandine is using newer one, from master branch. So you have to update precompiles to latest version (like with risc0, sp1 and ziren in [grandinetech/universal-precompiles](https://github.com/grandinetech/universal-precompiles) repo).

Also, since bls12_381 is currently set into `Cargo.toml`, does main grandine binary compile now?

rust-toolchain.toml
```diff
@@ -1,5 +1,8 @@
 [toolchain]
-channel = 'stable-2025-06-26'
+# channel = 'stable-2025-06-26'
+# Need nightly rust toolchain to build zkvm_host with `pico` feature
+channel = 'nightly-2025-08-04'
```
ArtiomTr: I personally don't think this is a good idea, but @povi & @sauliusgrigaitis know better, maybe they're okay with that.

However, is it possible to avoid somehow? Maybe we can put pico host code into separate crate, outside of main workspace, and compile it separately, with separate toolchain?

ssz/src/shared.rs
```diff
@@ -124,7 +124,8 @@ pub fn read_list<C, T: SszRead<C>, L: TryFromIterator<T, Error: Into<ReadError>>
         // TODO(feature/optimize-ssz-decoding): Benchmark without the early validation if that makes sense.
         let actual = bytes.len() / size;
 
-        if actual > maximum {
+        // Note: ignore the check `actual > maximum` when maximum == 0. Is this proper?
+        if maximum > 0 && actual > maximum {
```
ArtiomTr: I see this as very dangerous check (as it is essentially skipping check). Would be better to find exact place where `List` has max length of zero, and see what's happening there?

Still this place needs to be reviewed by someone who works with ssz more closely, maybe @povi can consult here?

zkvm/host/src/backend/pico/build_helper.rs
```diff
@@ -0,0 +1,31 @@
+use anyhow::{Result, Context};
+use std::{env, process::Command, path::Path};
+
+const GUEST_PROGRAM: &str = "zkvm-guest-pico.elf";
+
+pub fn build_program(path: &str) -> Result<()> {
```
ArtiomTr: Why this helper lives in its own crate? Maybe you can move it to `backend/pico`?

zkvm/host/src/backend/pico.rs
```diff
@@ -0,0 +1,157 @@
...
+        stdin_builder.write_slice(&state_ssz);
+        stdin_builder.write_slice(&block_ssz);
+        stdin_builder.write_slice(&cache_ssz);
+        stdin_builder.write_slice(&phase_bytes);
+
+        println!("zkvm_host: proving starts.");
+
+        let proof_set = prover
+            .prove(stdin_builder)
+            .expect("Failed to generate proof");
```
ArtiomTr: It is better to not panic here, but return error

zkvm/host/src/backend/pico.rs
```diff
@@ -0,0 +1,157 @@
...
+    ) -> Result<(Vec<u8>, Self::Report)> {
+        // Load the ELF file
+        let out_dir = env!("OUT_DIR");
+        let zkvm_guest_path = Path::new(&out_dir).join(ZKVM_GUEST_PICO_SUFFIX);
+        let elf = fs::read(&zkvm_guest_path).map_err(|err| {
+            anyhow!(format!(
+                "Failed to load ELF file from {}: {}",
+                zkvm_guest_path.display(),
+                err
+            ))
```
ArtiomTr: I believe this may be simplified with [anyhow's context feature](https://docs.rs/anyhow/1.0.100/anyhow/trait.Context.html#tymethod.with_context)? It should keep original error, while providing new error message (like saying "Failed to load ELF file from {some path}", and then showing e.g. "File not found" error below)

ArtiomTr (changes requested):

zkvm/host/src/backend/pico/build_helper.rs
jimmychu0807: Sure. Can move this in `backend/pico`.

rust-toolchain.toml
jimmychu0807: How about I keep the stable version line, and add a comment below for those who want to compile zkvm pico? Essentially:

```toml
channel = 'stable-2025-06-26'
# To buld zkvm_host with `pico` feature, comment the above and use the following nighly version.
# channel = 'nightly-2025-08-04'
```

Cargo.toml
jimmychu0807: This is what I'm not sure of. Brevis Pico has their own tweak of `bls12_381` and `sha2`. How can I make these crates compatible with other zkvms? (Actually each zkvm seems to have their own tweak on these two libs)

bls/bls-zkcrypto/Cargo.toml
jimmychu0807: I thought so at the beginning. For pico, the target value they use is **riscv32im-risc0-zkvm-elf**, which is very similar to r0vm.

rust-toolchain.toml
ArtiomTr: That works too, but adds some tedious work to run tests, which can be automated. Also, one more idea just came into my mind: you can use `cargo +nightly build` to build zkvm host with nightly version. So you can add target to makefile, to avoid manually commenting lines

ssz/src/shared.rs
jimmychu0807: @povi would like to add more context here. Without this fix (`maximum > 0` and also below), even the simplest test `consensus spec tests mainnet electra empty block transition` will fail inside the guest code on the line:

```rs
BeaconState::<P>::from_ssz_at_phase(phase, &state_ssz)
```
The list `maximum` passed in is 0, but there are content inside.

After reading this comment, if you also think there is a better approach, let me know and I will revert this fix in this PR and let you send a proper fix to `ssz` crate.

rust-toolchain.toml
jimmychu0807: Fixed, by restoring back to stable version. Added  lines in `Makefile` with `cargo +nightly-<version> build` to build zkvm-pico host and guest code.

zkvm/host/src/backend/pico/build_helper.rs
jimmychu0807: Done. Moved the function to `backend/pico/build_helper.rs`.

zkvm/host/src/backend/pico.rs
jimmychu0807: fixed

zkvm/host/src/backend/pico.rs
jimmychu0807: fixed

Cargo.toml
jimmychu0807: Pointed to 
`git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'RustCrypto-hashes/sha2-v0.10.9'` now.
Once https://github.com/grandinetech/universal-precompiles/pull/3 is merged, I will point back to `grandine/universal-precompiles.git`.

ssz/src/shared.rs
povi: No, don't remove the check. If the maximum is 0, that most likely means that `BeaconState` is from the different phase than it tries to decode from. Please re-check from which phase is `BeaconState` and what is the `phase` parameter that is being used in `from_ssz_at_phase`

jimmychu0807: I will restore this file to its previous state. When this PR is completed & merged, I will try to look into this ssz issue.

jimmychu0807: restored.

jimmychu0807:
latest CI result can be seen here:
https://github.com/jimmychu0807/grandine/actions

bls/bls-zkcrypto/src/secret_key.rs
jimmychu0807: It passed all CI tests: https://github.com/jimmychu0807/grandine/actions/runs/18261651253

Cargo.toml
```diff
@@ -299,7 +304,9 @@ base64 = '0.22'
 bincode = '1'
 bit_field = '0.10'
 bitvec = '1'
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+# bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+bls12_381 = { git = "https://github.com/jimmychu0807/grandine-universal-precompiles.git", branch = "zkcrypto/bls12_381" }
```
jimmychu0807, line 308: Once https://github.com/grandinetech/universal-precompiles/pull/3 is merged, modify to

```suggestion
bls12_381 = { git = "https://github.com/grandine/universal-precompiles.git", branch = "zkcrypto/bls12_381" }
```

jimmychu0807: Once https://github.com/grandinetech/universal-precompiles/pull/4 is merged, modify to

```suggestion
bls12_381 = { git = "https://github.com/grandine/universal-precompiles.git", branch = "zkcrypto/bls12_381" }
```

Cargo.toml
```diff
@@ -604,7 +613,9 @@ lto = false
...
 # We had to fork `jsonrpc` because it does not allow nonstandard members.
 jsonrpc-core = { git = 'https://github.com/grandinetech/jsonrpc.git' }
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
+# sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
+sha2 = { git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'RustCrypto-hashes/sha2-v0.10.9' }
 
 [patch.'https://github.com/zkcrypto/bls12_381.git']
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+# bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
+bls12_381 = { git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'zkcrypto/bls12_381' }
```
jimmychu0807: Once https://github.com/grandinetech/universal-precompiles/pull/4 is merged, modify to

```suggestion
bls12_381 = { git = "https://github.com/grandine/universal-precompiles.git", branch = "zkcrypto/bls12_381" }
```

Cargo.toml
```diff
@@ -433,7 +440,9 @@ serde_with = '3'
 serde_yaml = '0.9'
 # The `asm` feature in `sha2` doesn't do anything on recent `x86_64` CPUs
 # because `sha2` defaults to using CPU intrinsics.
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0', features = ['compress'] }
+# sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0', features = ['compress'] }
+sha2 = { git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'RustCrypto-hashes/sha2-v0.10.9', features = ['compress'] }
```
jimmychu0807: Once https://github.com/grandinetech/universal-precompiles/pull/3 is merged, modify to

```suggestion
sha2 = { git = "https://github.com/grandine/universal-precompiles.git", branch = "RustCrypto-hashes/sha2-v0.10.9" }
```

Cargo.toml
```diff
@@ -604,7 +613,9 @@ lto = false
 # `geth` responds to invalid payloads with objects containing `method` and `params`.
 # We had to fork `jsonrpc` because it does not allow nonstandard members.
 jsonrpc-core = { git = 'https://github.com/grandinetech/jsonrpc.git' }
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
+# sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
+sha2 = { git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'RustCrypto-hashes/sha2-v0.10.9' }
```
jimmychu0807: Once https://github.com/grandinetech/universal-precompiles/pull/3 is merged, modify to

```suggestion
sha2 = { git = "https://github.com/grandine/universal-precompiles.git", branch = "RustCrypto-hashes/sha2-v0.10.9" }
```

ssz/src/shared.rs
jimmychu0807: This line is remaining here, so the [zkvm test](https://github.com/jimmychu0807/grandine/actions/runs/18798125523) could pass. Let me know if you feel it is better left off. I will revert the fix.

povi:
@jimmychu0807, I see build fails (https://github.com/jimmychu0807/grandine/commit/afb0dace47fd363fec4d3e4fc8191e70944e9449). Please make sure CI checks pass

bls/bls-zkcrypto/src/secret_key.rs
ArtiomTr: This is still not fixed, there should be single api for `hash_to_curve`, not dependent on the backend

ssz/src/shared.rs
ArtiomTr: This should be removed, maximum check is required - error is caused because state gets deserialized with incorrect phase

ArtiomTr (changes requested):

ssz/src/shared.rs
jimmychu0807: reverted

jimmychu0807: removed

jimmychu0807:
@ArtiomTr

- ssz-check is reverted back
- build now succeeded, [see here](https://github.com/jimmychu0807/grandine/actions/runs/18974739899).
- you will see [zkvm build test fails](https://github.com/jimmychu0807/grandine/actions/runs/18974739888) because the ssz-check is reverted back.
  Created an issue on this: https://github.com/grandinetech/grandine/issues/454
- [bls/bls-zkcrypto/src/secret_key.rs](https://github.com/grandinetech/grandine/pull/386/files/99818d033464ade8170b43cea7c97940be531a79#diff-bbad1fc544ca0bfa7f5dc97fac2339e2f3f233cb456210f00143e0f3244251af) has a unified interface now given the bls12_381 update.

I believe the PR is now good for review. Thanks.

.github/workflows/ci.yml
```diff
@@ -27,7 +27,8 @@ jobs:
         if: runner.os == 'Linux'
         uses: easimon/maximize-build-space@master
         with:
-          root-reserve-mb: 2048
+          # Leaving 20 GB for rust crates download and installation under `/tmp`
+          root-reserve-mb: 20480
```
jimmychu0807: I encountered running out of space error because rust crates are downloaded under `/tmp` dir for installation. So here, we increase the space reserved for root dir.

Cargo.toml
```diff
@@ -102,7 +107,7 @@ unsafe_code = 'forbid'
 absolute_paths_not_starting_with_crate = 'warn'
 anonymous_parameters                   = 'warn'
 deprecated_in_future                   = 'warn'
-deprecated_safe                        = 'warn'
+deprecated_safe                        = { level = 'warn', priority = -1 }
```
ArtiomTr: why this change?

zkvm/guest/risc0/guest/Cargo.toml
```diff
@@ -18,9 +21,5 @@ ssz = { path = '../../../../ssz' }
 transition_functions = { path = '../../../../transition_functions' }
 types = { path = '../../../../types' }
 
-[patch.crates-io]
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
-
-[patch.'https://github.com/zkcrypto/bls12_381.git']
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
-
```
ArtiomTr, line 26: why bls patch removed?

zkvm/guest/sp1/Cargo.toml
```diff
@@ -14,9 +14,3 @@ sp1-zkvm = { version = '5.2.1', features = ['embedded'] }
 ssz = { path = '../../../ssz' }
 transition_functions = { path = '../../../transition_functions' }
 types = { path = '../../../types' }
-
-[patch.crates-io]
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.0' }
-
-[patch.'https://github.com/zkcrypto/bls12_381.git']
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.0' }
```
ArtiomTr, line 22: why these removed?

zkvm/host/Cargo.toml
```diff
@@ -25,17 +28,30 @@ types = { workspace = true }
 xz2 = { workspace = true }
 zkvm_guest_risc0 = { workspace = true, optional = true }
 
+# For pico zkvm
+pico-sdk = { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" , optional = true }
+pico-vm =  { git = "https://github.com/brevis-network/pico.git", tag = "v1.1.7" , optional = true }
+
 [build-dependencies]
 sp1-helper = { workspace = true, optional = true }
+anyhow = { workspace = true, optional = true }
```
ArtiomTr: why anyhow in build dependencies?

.github/workflows/zkvm-test.yml
```diff
@@ -0,0 +1,89 @@
...
+
+jobs:
+  zkvm-test:
+    strategy:
+      matrix:
+        test_case: [
+          # These two test cases can be completed in a reasonably short time
+          "consensus spec tests mainnet electra empty block transition",
+          "pectra-devnet-6 without epoch transition"
+        ]
```
ArtiomTr: I think it is better to structure pipeline the other way - by doing matrix on "backend", not on test case. Then you won't need to install each backend twice.

.github/workflows/zkvm-test.yml
```diff
@@ -0,0 +1,89 @@
...
+        lfs: true
+
+    - name: Check out necessary submodules only
+      run: |
+        git submodule update --init dedicated_executor eth2_libp2p
+
+    - name: Installing r0vm
+      run: |
+        curl -L https://risczero.com/install | bash
+        ls -lah $HOME
```
ArtiomTr: is this really needed?

.github/workflows/zkvm-test.yml
```diff
@@ -0,0 +1,89 @@
...
+        source /home/runner/.bashrc
+        /home/runner/.risc0/bin/rzup install rust 1.88.0
+        /home/runner/.risc0/bin/rzup install cpp 2024.1.5
+        /home/runner/.risc0/bin/rzup install r0vm 3.0.1
+        /home/runner/.risc0/bin/rzup install cargo-risczero 3.0.1
```
ArtiomTr: shouldn't it work by providing not whole path, but just executable name `rzup`?

.github/workflows/zkvm-test.yml
```diff
@@ -0,0 +1,89 @@
...
+      run: |
+        rustup install nightly-2025-08-04
+        rustup component add rust-src --toolchain nightly-2025-08-04
+        pushd ./
+        cd ..
+        git clone https://github.com/brevis-network/pico pico
+        cd pico/sdk/cli
+        cargo install --locked --force --path .
+        cargo pico --version
+        popd
```
ArtiomTr: I believe all of this can be simplified to
```
cargo +nightly-2025-08-04 install --git https://github.com/brevis-network/pico pico-cli
```
as mentioned [in their docs](https://pico-docs.brevis.network/getting-started/installation.html#option-1--cargo-install)

Cargo.toml
```diff
@@ -68,6 +68,11 @@ members = [
     'zkvm/guest/risc0',
     'zkvm/host',
 ]
+exclude = [
+    # Excluding the zkvm guest library will make the compiled guest code size a lot smaller, from
+    # ~200 MB -> 2.3 MB.
+    'zkvm/guest/pico'
+]
```
ArtiomTr: cool, is it default exclude flag value? If so, we can simplify our makefile by putting excludes here:

https://github.com/grandinetech/grandine/blob/e53a535e114c55678207b9661adec44d8be2348b/Makefile#L1

ArtiomTr (changes requested):
Also please add documentation in `zkvm/README.md` on how to run pico tests.

ArtiomTr:
Also, can you please squash all commits into one, to keep history clean? Thank you

.github/workflows/zkvm-test.yml
ArtiomTr: also, after thinking a bit more about this, maybe this job should be put inside `./.github/workflows/zkvm-test.yml` workflow, rather than separate one? I believe two jobs - default lint/build/format CI and zkvm testing - should run at the same events, so we can colocate them properly

zkvm/guest/risc0/guest/Cargo.toml
jimmychu0807: We don't need this line to compile the guest, as shown in the workflows/zkvm-test.yml

ArtiomTr: ah yes, because of feature thing, okay

zkvm/guest/sp1/Cargo.toml
jimmychu0807: The guest compiles fine without these two lines

ArtiomTr: yes, but guest also compiles without any patch at all - the point of patches is not to provide functionality that otherwise is unavailable, but to speed up existing crates. By removing these sections, performance should degrade.

.github/workflows/zkvm-test.yml
jimmychu0807: `./.github/workflows/zkvm-test.yml`

Do you intent to mean `./.github/workflows/ci.yml`? It is now in `zkvm-test.yml`.

jimmychu0807: I think it will be better to have all zkVM related build/tests separated from the main ones, the `ci.yml` build/tests.

Cargo.toml
jimmychu0807: Since 'zkvm/guest/pico' is already excluded from workspace, there is no `zkvm_guest_pico` from the "workspace" perspective. Isn't it?

.github/workflows/zkvm-test.yml
jimmychu0807: Refer to docs:
https://dev.risczero.com/api/zkvm/install#using-rzup-in-ci

.github/workflows/zkvm-test.yml
ArtiomTr: yeah, sorry, meant `ci.yaml`. I proposed to keep them joint, as that is still CI, checking if code works, not separate pipeline. They both need to run at the same time, so don't see a reason to keep them separate

.github/workflows/zkvm-test.yml
ArtiomTr: still doesn't answer my question, why you use `/home/runner/.risc0/bin/rzup` instead of simple `rzup`? Docs you pointed also show `rzup` usage, not full path

.github/workflows/zkvm-test.yml
jimmychu0807: > I think it is better to structure pipeline the other way - by doing matrix on "backend", not on test case. Then you won't need to install each backend twice.

I agree this can be structured to parallelize run for different zkVMs.

Cargo.toml
ArtiomTr: okay then I have zero clue what's going here - how you exclude package from workspace, that isn't there in the first place? https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

.github/workflows/zkvm-test.yml
jimmychu0807: It couldn't find `rzup` (not in the path). Shown here:
https://github.com/jimmychu0807/grandine/actions/runs/19332684742/job/55299564273#step:2:41

So using the full path here.
Updated: at the end I added an `export PATH...` command so we can run `rzup` directly. But it is not in the default PATH even after sourcing `~/.bashrc`.

Cargo.toml
jimmychu0807: The **exclude** inside the `Cargo.toml` is excluding the specified child path to be included as a package.

The **EXCLUDES** that we set in Makefile is excluding whatever packages already recognized by the workspace, thus using the crate name.

Makefile
```diff
@@ -5,6 +5,7 @@ check:
 	cargo check --features default-networks $(EXCLUDES)
 
 build:
+	cargo build --release --bin grandine --features default-networks --workspace --exclude zkvm_host --exclude zkvm_guest_risc0
```
ArtiomTr: I believe this line not needed

.github/workflows/ci.yml
```diff
@@ -75,3 +80,88 @@ jobs:
...
+        rustup component add rust-src --toolchain nightly-2025-08-04
+        cargo +nightly-2025-08-04 install --git https://github.com/brevis-network/pico pico-cli
+
+    - name: Running Brevis Pico test
+      run: |
+        IFS=$'\n'
+        for TEST in $TEST_CASES; do
+          echo "Running test: $TEST"
+          RUST_LOG=info cargo +nightly-2025-08-04 run -p zkvm_host --features pico --release -- --test "$TEST" execute
+        done
```
ArtiomTr: still believe this all can be done more cleanly with matrix strategy, something like:

```yml
matrix:
  backend: [risc0, sp1, pico]
  include:
    - backend: risc0
      install: |
        curl -L https://risczero.com/install | bash
        source /home/runner/.bashrc
        export PATH="/home/runner/.risc0/bin:$PATH"
        rzup install rust 1.88.0
        rzup install cpp 2024.1.5
        rzup install r0vm 3.0.1
        rzup install cargo-risczero 3.0.1
        cargo risczero --version
      execute: RUST_LOG=info RISC0_DEV_MODE=1 cargo run -p zkvm_host --features risc0 --release -- --test "$TEST" execute
    
    - backend: sp1
      install: ...
      execute: ...
      
    - backend: pico
      install: ...
      execute: ...
      
steps:
  # preparation steps here
  
  - name: Installing ${{ matrix.backend }} zkvm
    run: ${{ matrix.install }}
    
  - name: Running ${{ matrix.backend }} tests
    run: |
      IFS=$'\n'
      for TEST in $TEST_CASES; do
        echo "Running test: $TEST"
        ${{ matrix.execute }}
      done
```
Then no need for setup phase in separate file - everything can be in one place. Also, simplifies adding new zkvm

Cargo.toml
ArtiomTr: okay, I checked locally, and found some things:
* The comment is misleading, as this exclude required because otherwise it will error out, as guest code believes that it lives in workspace:
  ```
    thread 'main' panicked at sdk/cli/src/build/build.rs:140:10:
    cargo metadata command failed: CargoMetadata { stderr: "error: current package believes it's in a workspace when it's not:\ncurrent:   /home/necolt/workspace/grandine/zkvm/guest/pico/Cargo.toml\nworkspace: /home/necolt/workspace/grandine/Cargo.toml\n\nthis may be fixable by adding `zkvm/guest/pico` to the `workspace.members` array of the manifest located at: /home/necolt/workspace/grandine/Cargo.toml\nAlternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.\n" }
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
  ```
* As suggested by error, two things can be done: either adding an exclude to workspace cargo.toml, or adding `[workspace]` section to guest's cargo.toml. 
* The same issue was with SP1's guest code, and it was solved by adding empty `[workspace]` section. 
* For the sake of consistency, please remove these excludes and add empty workspace section to pico's Cargo.toml. 
* Double check guest code size: if empty `[workspace]` section produces larger guest code, than `[excludes]` field, then remove empty workspace sections in both sp1 and pico guest code.

zkvm/README.md
```diff
@@ -1,24 +1,75 @@
...
+   Compiling Brevis pico libraries require using nightly rust. Prefix command with `RUST_LOG=info` to include enhanced usage stats.
+
+2. **Prove using local prover:**
+
+   ```sh
+   cargo +nightly-2025-08-04 run -p zkvm_host --features pico --release -- --test <test_case> prove
+   ```
+   Again, replace `<test_case>` with one of the four test cases mentioned above.
 
 ## Adding new zkvm
```
ArtiomTr, line 74: Can you please also mention in `Adding new zkvm` section, that you need to add new ZKVM to ci.yaml workflow?

ArtiomTr (changes requested):

Cargo.toml
jimmychu0807: added an empty [workspace] and removed the [excludes] field. The output elf file size is the same.

jimmychu0807:
I didn't change the disk-space-freer. I tried `jlumbroso/free-disk-space` and it gave other errors.

So sticking with `easimon/maximize-build-space@master` for now.  When the CI is complete:
https://github.com/jimmychu0807/grandine/actions/runs/19362033854

See if the PR is good to merge. Thanks.

ArtiomTr:
> I didn't change the disk-space-freer. I tried `jlumbroso/free-disk-space` and it gave other errors.

That's because it uninstalls too much, and now required dependencies for grandine must be installed back:

```yaml
      - name: Install system dependencies (Ubuntu)
        if: runner.os == 'Linux'
        run: |
          sudo DEBIAN_FRONTEND='noninteractive' apt-get install \
            --no-install-recommends -yq \
            ca-certificates \
            libssl-dev \
            clang \
            cmake \
            unzip \
            protobuf-compiler \
            libz-dev
```

> See if the PR is good to merge. Thanks.

The flaky CI issue is quite a big deal, needs to be sorted out before merging this PR.

jimmychu0807:
But this CI has been used in Grandine since I started working on this. I'm not sure I still have time to work on this CI issue. If not, I would suggest fixing this in another PR.

It passes (this time):
https://github.com/jimmychu0807/grandine/actions/runs/19362033854/job/55398968665

jimmychu0807:
@ArtiomTr, I have updated `ci.yaml` using **jlumbroso/free-disk-space@main** action and added the section of code to add back system dependencies.

The CI is running fine for the last push:
https://github.com/jimmychu0807/grandine/actions/runs/19460867176

ArtiomTr (approved):
LGTM :+1:

jimmychu0807:
@ArtiomTr see if this PR is good to merge. Thanks.

jimmychu0807:
Hi @povi, this is the integration of Brevis Pico in the Grandine zkvm crate (both host and guest). Please see if you have time to review and get it merged to the `develop` branch. Thank you!

povi (approved):
Looks good, let's merge

PR #399, opened 2025-10-03
Add Gloas helper and state transition functions

### LINKED ISSUE #417: Add Gloas helper, state transition functions
spec: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#helper-functions

hangleang:


eip_7594/src/lib.rs
```diff
@@ -294,7 +290,52 @@ fn get_data_column_sidecars<P: Preset>(
...
+        )?;
+
+        sidecars.push(Arc::new(
+            GloasDataColumnSidecar {
+                index: ColumnIndex::try_from(column_index)?,
+                column,
+                kzg_commitments: kzg_commitments.clone(),
+                kzg_proofs,
+                beacon_block_root,
+                slot,
```
povi, line 334: FYI `beacon_block_root` and `slot` got swapped in places: https://github.com/ethereum/consensus-specs/blob/v1.6.0/specs/gloas/p2p-interface.md#modified-datacolumnsidecar

fork_choice_control/src/mutator.rs
```diff
@@ -1559,15 +1559,22 @@ where
...
-                    data_column_sidecar.signed_block_header.message,
-                    data_column_sidecar.index,
-                ) {
-                    let block_root = data_column_sidecar
-                        .signed_block_header
-                        .message
-                        .hash_tree_root();
+                let block_root = data_column_sidecar.beacon_block_root();
+                let is_accepted_sidecar =
+                    if let Some(data_column_sidecar) = data_column_sidecar.pre_gloas() {
```
povi: This could be handled internally in `Store::accepted_data_column_sidecar` method. Then it'd be no need to do the same check all over the place.

types/src/combined.rs
```diff
@@ -412,6 +415,20 @@ impl<P: Preset> BeaconState<P> {
...
+    pub fn post_gloas(&self) -> Option<&dyn PostGloasBeaconState<P>> {
+        match self {
+            Self::Phase0(_)
+            | Self::Altair(_)
+            | Self::Bellatrix(_)
+            | Self::Capella(_)
+            | Self::Deneb(_)
+            | Self::Electra(_)
+            | Self::Fulu(_) => None,
+            Self::Gloas(_state) => None,
```
povi: Is this line correct?

hangleang: how can I miss this, thank you

hangleang: apply those suggestions/fix in https://github.com/grandinetech/grandine/pull/399/commits/a3091a1f9f32435fa1d7976102e7fe11f766a30d

povi (approved):

PR #406, opened 2025-10-08
Compute blob poofs post fulu for `/eth/v1/beacon/blob_sidecars/{block_id}`

### LINKED ISSUE #405: Beacon API: `/eth/v1/beacon/blob_sidecars` compute blob proofs from blobs and commitments

hangleang:
close #405

### DIFF
--- a/http_api/src/standard.rs
+++ b/http_api/src/standard.rs
@@ -34,6 +34,7 @@ use genesis::AnchorCheckpointProvider;
 use helper_functions::{accessors, misc};
 use http_api_utils::{BlockId, StateId};
 use itertools::{izip, Either, Itertools as _};
+use kzg_utils::eip_4844::compute_blob_kzg_proof;
 use liveness_tracker::ApiToLiveness;
 use log::{debug, info, warn};
 use operation_pools::{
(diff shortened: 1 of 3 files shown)


http_api/src/standard.rs
```diff
@@ -1240,17 +1241,22 @@ pub async fn blob_sidecars<P: Preset, W: Wait>(
...
             block.clone_arc(),
             block_root,
             metrics.as_ref(),
         )
         .await?;
+        let blob_proofs = blobs
+            .iter()
+            .zip(commitments.into_iter())
+            .map(|(blob, commitment)| {
+                compute_blob_kzg_proof::<P>(blob, commitment, controller.store_config().kzg_backend)
```
povi: Looks good, except this part: `compute_blob_kzg_proof` is heavy cpu-blocking function, it blocks `tokio` executor. As a result, while proofs are computed, HTTP API probably won't handle other requests.

What you should do here is to move all blob_proofs computation to a single `tokio::task::spawn_blocking` task (alternatively you could use `tokio::task::block_in_place`)

hangleang: will do that

povi (approved):
LGTM

PR #413, opened 2025-10-09
Check data availability ranges in range requests

Tumas:


p2p/src/network.rs
```diff
@@ -1312,6 +1313,26 @@ impl<P: Preset> Network<P> {
...
+            debug!(
+                "received invalid request requesting data before availability period: \
+                (peer_id: {peer_id}, inbound_request_id: {inbound_request_id:?}, \
+                request: {request:?})",
+            );
+
+            ServiceInboundMessage::SendErrorResponse(
+                peer_id,
+                inbound_request_id,
+                RpcErrorResponse::InvalidRequest,
```
hangleang, line 1322:
```suggestion
                RpcErrorResponse::ResourceUnavailable,
```
should respond with `3: ResourceUnavailable`? https://github.com/ethereum/consensus-specs/blob/44db4a3aa18db12a16b09f4b808b656427825974/specs/deneb/p2p-interface.md?plain=1#L402

p2p/src/network.rs
```diff
@@ -1375,16 +1396,32 @@ impl<P: Preset> Network<P> {
...
```
hangleang, line 1405:
```suggestion
                RpcErrorResponse::ResourceUnavailable,
```
same as above

hangleang (commented):
.

p2p/src/network.rs
Tumas: This states that `ResourceUnavailable` be set for data that is is within the range
`blob_serve_range`. The if statement in the PR checks for requests that are asking data prior to `blob_serve_range`

hangleang: ahh I see, so invalid request here is correct. but we don't send `ResourceUnavailable` error in case of request within the serve range?

Tumas: no, at the moment we don't send anything for data we don't have

povi (approved):

Issue #14, opened 2024-03-14
Change tagline to from eth2 to CL


taxmeifyoucan:
The current repository description describes the project as 'The Fastest Ethereum 2.0 Client' which is using old eth2 nomenclature and can be quite confusing. The correct naming should be a 'Consensus client' or 'Consensus layer client'. This is the most significant reference I noticed so far but if there are others, consider updating it too. 

Btw congrats on going FOSS with GPL!

sauliusgrigaitis:
Fixed, thanks :)

Issue #11, opened 2024-01-27
Feature: Support Doppelganger detection on VC


yorickdowne:
Support doppelganger detection in some fashion for keys loaded into Grandine.

Definitely not urgent; could make sense after a standalone VC exists and is stable.

sauliusgrigaitis:
We probably will get this feature before standalone VC. Standalone VC will take some time really.

Tumas:
Hey, @yorickdowne  Doppelganger support has been added recently with 92e08abafb08feb733f094e0fcfaa6d694db68e9

Tumas:
I’m closing this issue as the requested functionality was implemented some time ago, and no further updates or activity have occurred since. If there are any additional concerns or related needs, please feel free to reopen this issue or create a new one.

PR #422, opened 2025-10-14
Full Repository Migration to Tokio Tracing & Dynamic Runtime Log Control

sntntn:
This PR merges and finalizes all previous logging-to-tracing efforts across the repository, completing  the **[Logging to Tracing migration Roadmap](https://github.com/eth-protocol-fellows/cohort-six/blob/master/projects/Grandine-Implementing-Tokio-Tracing-For-Debugging-And-Performance-Analysis.md#roadmap)** + **Dynamic Runtime Log Control trough API**. 
The migration is coordinated with the [PR in eth2_libp2p submodule](https://github.com/grandinetech/eth2_libp2p/pull/21).

Summary of Related PRs:
- [Tracing Migration – Foundational Setup](https://github.com/sntntn/grandine/pull/1)
- [Repository-Wide Tracing Migration & Full Network Layer Instrumentation](https://github.com/sntntn/grandine/pull/9)
- [Add runtime API endpoint for dynamic tracing log level control](https://github.com/sntntn/grandine/pull/18)

---

#### Simplified Logging Flow
The legacy `slog` bridge has been removed.
Submodules (e.g. `eth2_libp2p`) can now log directly through the global Tokio Tracing subscriber, without passing logger instances.

#### Custom Tracing Macros
New helper macros wrap the `PEER_LOG_METRICS` context into all standard log levels, ensuring that peer connection state is always visible in log output:
``` rust
info_with_peers!()
warn_with_peers!()
debug_with_peers!()
trace_with_peers!()
crit!() // The crit! macro is a specialized variant of error!
```
> Each macro expands to a standard tracing call with an additional structured field named `peers`

#### Instrumentation
Instrumentation improve observability and debuggability across async boundaries.

- **Function-Level Instrumentation** (`#[instrument]`)
    - Key async and control-flow functions are annotated with the `#[instrument]` macro.
    - This automatically creates structured spans capturing:
        - function entry and exit,
        - key parameters (e.g. slot ...),
        - execution timing and outcome.
    - Trace-level spans:
        - **Network layer** (*eth2_libp2p*): RPC, gossip paths..
        - **fork_choice_control**
        - **validator**
        

- **Async Flow Instrumentation** (`.instrument(info_span!(...))`)
    - Long-running or chained async tasks are wrapped in spans using the `.instrument()` method together with `info_span!` -> provides visibility into async lifetimes
    - Implemented primarily in *network tests*

>Note: Manual span management (e.g. calling .entered() inside async code) is not recommended — it can introduce **non-Send guards** and thread-safety issues. Instead, prefer **#[instrument]** or **.instrument(span)** for safe and structured span propagation. See: [Network layer async functions not Send (runtime + main affected, thread-safety issue) #14](https://github.com/sntntn/grandine/issues/14)

- **Span Context & Performance**
    - Each span propagates structured context (e.g. slot, service, name..).
    - Root spans are declared with `parent = None` for isolated operations.
    - Overhead is negligible when tracing at lower levels — spans are lazily constructed and emitted only when enabled.

---

#### Environment Variable Behavior (unchanged from legacy logger)
Program behavior via environment variables remains identical to the previous logger implementation.

A detailed explanation can be found in [issue #11 comment](https://github.com/sntntn/grandine/issues/11#issuecomment-3236070219)

- Logging levels can be configured via the environment variables:
    - **`GRANDINE_LOG`** – primary
    - **`RUST_LOG`** – fallback
- Supports per-module level configuration, for example:
``` rust
GRANDINE_LOG=info,pubkey_cache=debug
```
> applies debug logging for pubkey_cache while keeping global level at info.

- If neither `GRANDINE_LOG` nor `RUST_LOG` is set and the binary is simply run,
the system falls back to hardcoded directives defined in `binary_utils/src/lib.rs`.
In that case, `.add_directive(LevelFilter::OFF.into())` is applied and external logs are suppressed unless explicitly whitelisted, just as before.

#### Dynamic Runtime Log Level Control
This feature adds the ability to change tracing log levels at runtime via an API,
powered by `tracing_subscriber::reload` layers

- Endpoint
``` api
POST /eth/v2/debug/tracing/log_level
```
- Example
``` api
curl -X POST http://localhost:5052/eth/v2/debug/tracing/log_level \
  -H "Content-Type: application/json" \
  -d '{
    "target": "fork_choice_control",
    "level": "debug"
  }'
```
> [Video demo](https://www.youtube.com/embed/c8RrdKiOQno) - in this video example, the log level for the `fork_choice_control` module is changed from `info` → `debug` → back to `info` using a simple **POST** request.  

#### TTY-Based ANSI Style Detection
- ANSI color styles are now automatically enabled only when output is a TTY.
- When output is redirected, styles are disabled by default.
- However, users can explicitly override this behavior via the logger-always-write-style

---
#### All existing tests pass with the new tracing setup.


grandine/src/main.rs
```diff
@@ -350,14 +353,38 @@ fn main() -> ExitCode {
...
+    //             eprintln!("failed to parse directive: {e}");
+    //         }
+    //     }
+    // }) {
+    //     eprintln!("failed to reload filter: {e}");
+    // }
+
     binary_utils::initialize_rayon()?;
 
+    info_with_peers!("Tracing started!");
```
Tumas: [**question**](https://conventionalcomments.org/): Should this perhaps be a debug-level statement? It might make sense to print it from `binary_utils::initialize_tracing_logger` instead. Also, note our convention of starting log messages with a lowercase letter.

grandine/src/main.rs
```diff
@@ -350,14 +353,38 @@ fn main() -> ExitCode {
...
+    // let x1 = "smth1";
+    // let x2 = "smth2";
+    // let x3 = "smth3";
+    // logging::info_with_peers!(field1=x1, field2=x2, "Tracing started! {}",x3);
```
Tumas: [**todo**](https://conventionalcomments.org/): These look like temporary comments — if so, they should be removed.

binary_utils/src/lib.rs
```diff
@@ -1,78 +1,105 @@
...
+    {
+        self.0.modify(f)
     }
+}
+
+pub fn initialize_tracing_logger(
+    module_path: &str,
+    always_write_style: bool,
+) -> Result<TracingHandle> {
+    let base_filter =
```
Tumas: [**issue**](https://conventionalcomments.org/): if **any values** are set in `GRANDINE_LOG`, they will overwrite **all** hardcoded filter values. For example, `GRANDINE_LOG=p2p=debug` will enable logging **only** for `p2p`. This is not backwards-compatible with current setup and could be addressed as suggested below, or you may find a more robust solution:

```diff
-    let base_filter =
-        EnvFilter::try_from_env("GRANDINE_LOG").or_else(|_| EnvFilter::try_from_default_env());
+    let mut filter = EnvFilter::default()
+        .add_directive(LevelFilter::OFF.into())
+        .add_directive("attestation_verifier=info".parse()?)
+        .add_directive("block_producer=info".parse()?)
+        .add_directive("builder_api=info".parse()?)
+        .add_directive("data_dumper=info".parse()?)
+        .add_directive("database=info".parse()?)
+        .add_directive("dedicated_executor=info".parse()?)
+        .add_directive("doppelganger_protection=info".parse()?)
+        .add_directive("eth1=info".parse()?)
+        .add_directive("eth1_api=info".parse()?)
+        .add_directive("eth2_libp2p=info".parse()?)
+        .add_directive("execution_engine=info".parse()?)
+        .add_directive("features=info".parse()?)
+        .add_directive("fork_choice_control=info".parse()?)
+        .add_directive("fork_choice_store=info".parse()?)
+        .add_directive("genesis=info".parse()?)
+        .add_directive("http_api=info".parse()?)
+        .add_directive("http_api_utils=info".parse()?)
+        .add_directive("keymanager=info".parse()?)
+        .add_directive("liveness_tracker=info".parse()?)
+        .add_directive("metrics=info".parse()?)
+        .add_directive("operation_pools=info".parse()?)
+        .add_directive("p2p=info".parse()?)
+        .add_directive("prometheus_metrics=info".parse()?)
+        .add_directive("pubkey_cache=info".parse()?)
+        .add_directive("runtime=info".parse()?)
+        .add_directive("signer=info".parse()?)
+        .add_directive("slasher=info".parse()?)
+        .add_directive("slashing_protection=info".parse()?)
+        .add_directive("state_cache=info".parse()?)
+        .add_directive("storage=info".parse()?)
+        .add_directive("validator=info".parse()?)
+        .add_directive("validator_key_cache=info".parse()?)
+        .add_directive("validator_statistics=info".parse()?)
+        .add_directive("web3=info".parse()?)
+        .add_directive(format!("{module_path}=info").parse()?)
+        .add_directive(format!("{}=info", module_path!()).parse()?);
 
-    let filter = match base_filter {
-        Ok(filter) => filter,
-        Err(_) => {
-            let filter = EnvFilter::default()
-                .add_directive(LevelFilter::OFF.into())
-                .add_directive("attestation_verifier=info".parse()?)
-                .add_directive("block_producer=info".parse()?)
-                .add_directive("builder_api=info".parse()?)
-                .add_directive("data_dumper=info".parse()?)
-                .add_directive("database=info".parse()?)
-                .add_directive("dedicated_executor=info".parse()?)
-                .add_directive("doppelganger_protection=info".parse()?)
-                .add_directive("eth1=info".parse()?)
-                .add_directive("eth1_api=info".parse()?)
-                .add_directive("eth2_libp2p=info".parse()?)
-                .add_directive("execution_engine=info".parse()?)
-                .add_directive("features=info".parse()?)
-                .add_directive("fork_choice_control=info".parse()?)
-                .add_directive("fork_choice_store=info".parse()?)
-                .add_directive("genesis=info".parse()?)
-                .add_directive("http_api=info".parse()?)
-                .add_directive("http_api_utils=info".parse()?)
-                .add_directive("keymanager=info".parse()?)
-                .add_directive("liveness_tracker=info".parse()?)
-                .add_directive("metrics=info".parse()?)
-                .add_directive("operation_pools=info".parse()?)
-                .add_directive("p2p=info".parse()?)
-                .add_directive("prometheus_metrics=info".parse()?)
-                .add_directive("pubkey_cache=info".parse()?)
-                .add_directive("runtime=info".parse()?)
-                .add_directive("signer=info".parse()?)
-                .add_directive("slasher=info".parse()?)
-                .add_directive("slashing_protection=info".parse()?)
-                .add_directive("state_cache=info".parse()?)
-                .add_directive("storage=info".parse()?)
-                .add_directive("validator=info".parse()?)
-                .add_directive("validator_key_cache=info".parse()?)
-                .add_directive("validator_statistics=info".parse()?)
-                .add_directive("web3=info".parse()?)
-                .add_directive(format!("{module_path}=info").parse()?)
-                .add_directive(format!("{}=info", module_path!()).parse()?);
-
-            filter
+    if let Ok(env_filter) = EnvFilter::try_from_env("GRANDINE_LOG") {
+        for directive in env_filter.to_string().split(',') {
+            filter = filter.add_directive(directive.parse()?)
         }
-    };
+    }
```

fork_choice_control/src/controller.rs
```diff
@@ -239,6 +240,15 @@ where
     // More or less frequent calls are allowed but may worsen performance and quality of the head.
     // According to the Fork Choice specification, `on_tick` should be called every second,
     // but doing so would be redundant. The fork choice rule does not need a precise timestamp.
+    #[instrument(
+        parent = None,
+        level = "trace",
+        fields(
+            service = "fork_choice"
+        ),
+        name = "fork_choice_controll",
```
Tumas: [**chore**](https://conventionalcomments.org/): Fix typo: replace `controll` with `control`. This typo appears in multiple places.

logging/src/lib.rs
```diff
@@ -30,3 +30,64 @@ impl PeerLogMetrics {
...
+            peers = %$crate::PEER_LOG_METRICS,
+            $( $rest )*
+        )
+    };
+}
+
+#[macro_export]
+macro_rules! crit {
+    ( $( $rest:tt )* ) => {
+        ::tracing::error!(
```
Tumas, line 87: [**question**](https://conventionalcomments.org/): Shouldn't `crit` messages be written to a dedicated log file rather than appearing in the general output? Currently, they are logged to the general output with a custom `error_type` field. 

As I understand it, `crit!` is meant primarily for developers and should be hidden from the main log view unless debug mode is enabled for that module. In other words, `crit!` should behave like a debug-level message when debug logging is active for the module, and otherwise be routed to the dedicated log file.

You can check the similar approach in lighthouse:
https://github.com/sigp/lighthouse/blob/stable/common/logging/src/tracing_logging_layer.rs

`custom crit-log` could be tackled in another pull request though.

http_api/src/task.rs
```diff
@@ -143,7 +150,7 @@ impl<P: Preset, W: Wait> HttpApi<P, W> {
 
         let handle_sync_statuses = handle_sync_statuses(is_synced, sync_to_api_rx);
 
-        info!("HTTP server listening on {address}");
+        info_with_peers!("HTTP server listening");
```
Tumas: [**question**](https://conventionalcomments.org/): any reason why `address` is no longer in the logs?

grandine/src/main.rs
sntntn: Sure, make sense

grandine/src/main.rs
sntntn: yes, my mis

http_api/src/task.rs
sntntn: My miss - the address was removed when I was experimenting with tracing fields. I’ll add it back.

binary_utils/src/lib.rs
sntntn: Yes, that was indeed the previous behavior - thanks for pointing it out!
I’ve updated it following your suggestion, and also added `EnvFilter::try_from_default_env()`
to keep the default tracing behavior through `RUST_LOG`.
It can easily be removed if we decide not to rely on it.

logging/src/lib.rs
sntntn: Correct, in the current implementation there's no redirection yet.
I do include the `error_type` field, so I can use that to route `crit!` logs separately while in info mode.
As I understand it, when `debug` mode is enabled for that module, you’d like `crit!` messages to also appear in the main log output (like error?, or like debug?).

I’ll open a follow-up PR to implement that as you suggested.

Tumas: > As I understand it, when debug mode is enabled for that module, you’d like crit! messages to also appear in the main log output (like error?, or like debug?).

Correct. But if that’s tricky to implement, we can just keep crit! messages only in a separate log instead.

sntntn: I managed to implement developer-controlled crit! logging across all log levels, so it can now function both in debug mode and in production without the need for a separate log file.

- Developers can now enable crit! logging on demand.
- crit! visibility can be controlled via:
    - Environment variable `GRANDINE_LOG` or `RUST_LOG`
    - Reloadable tracing layer / API (`POST /eth/v2/debug/tracing/log_level`) - so logging can be enabled dynamically after startup.
- No dedicated log file is required; logs are filtered by the crit target.

full implementation in PR: [#432](https://github.com/grandinetech/grandine/pull/432)

PR #425, opened 2025-10-15
Ensure data column sidecars respect the blob limit

hangleang:
spec change: https://github.com/ethereum/consensus-specs/pull/4650

### DIFF
--- a/eip_7594/src/lib.rs
+++ b/eip_7594/src/lib.rs
@@ -144,12 +144,16 @@ pub fn compute_subnets_for_node<P: Preset>(
 }
 
 /// Verify if the data column sidecar is valid.
-pub fn verify_data_column_sidecar<P: Preset>(data_column_sidecar: &DataColumnSidecar<P>) -> bool {
+pub fn verify_data_column_sidecar<P: Preset>(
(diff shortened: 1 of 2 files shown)


eip_7594/src/lib.rs
```diff
@@ -144,12 +144,16 @@ pub fn compute_subnets_for_node<P: Preset>(
 }
 
 /// Verify if the data column sidecar is valid.
-pub fn verify_data_column_sidecar<P: Preset>(data_column_sidecar: &DataColumnSidecar<P>) -> bool {
+pub fn verify_data_column_sidecar<P: Preset>(
+    data_column_sidecar: &DataColumnSidecar<P>,
```
Tumas, line 148: [**nitpick**](https://conventionalcomments.org/): We usually follow the convention (though not strictly enforced) of placing Config as the first argument.

Tumas (approved):

eip_7594/src/lib.rs
hangleang: ohh yeah, it has been put incorrect order for the whole module, would need another PR to reorder all those functions

Issue #455, opened 2025-11-01
exeption.log file rotation after reaching 1GB


sntntn:

Tumas:
Closing as complete

Issue #451, opened 2025-10-30
Fix exeption.log file path


sauliusgrigaitis:
@sntntn `exception.log` file by default must be written to `<data-dir>\<network>\exception.log`

sauliusgrigaitis:
Also, now seems that Grandine just exits silently if it can't write that `exception.log` file. It needs to fail with error message and stop the client if this `exception.log` file is mandatory or alternatively just a warning and proceed with execution.

sntntn:
Makes sense. I'll do it today!

sauliusgrigaitis:
> Makes sense. I'll do it today!

@povi  is already working on it.

I suggest to explore the metrics as we talked before.

povi:
> > Makes sense. I'll do it today!
> 
> [@povilas](https://github.com/povilas) is already working on it.
> 
> I suggest to explore the metrics as we talked before.

I think you referenced a wrong person 😃

povilas:
> [@povilas](https://github.com/povilas) is already working on it.

fake news!

sntntn:
excellent, then I'll move on metrics research :D
I suppose @povi would also rotate the exception.log file after exceeding memory limit then

PR #456, opened 2025-11-01
rotate exception.log when it exceeds 1GB

sntntn:
Currently, exception.log could grow indefinitely, eventually(worst case) consuming all available disk space.

This PR migrates the exception logging system to use the logroller crate, introducing size-based rotation, compression, and file retention:

- Rotation strategy: size-based, with a 1 GB limit per file
- Compression: old logs are automatically compressed using gzip
- Retention: up to 5 log files are kept (1 current + 4 compressed)

In the worst case, this setup can retain up to ~2 GB of exception logs on disk,
which would correspond to about 5 GB of uncompressed data.
> compressed logs reduce their size by approximately 5–6×

---

solves #455  (continues the conversation from #451)

---

## Verification:

The rotation and compression behavior were verified during runtime by **temporarily lowering** the rotation threshold to **1 KB**.

This allowed observing rotations and compressed files being correctly generated and retained.
An exception message was injected in one of the **frequently** executed parts of the code before node startup (specifically in `execution_service.rs`).

test example:

Screenshot from 2025-11-03 19-50-16

Screenshot from 2025-11-03 19-50-31

### DIFF
--- a/binary_utils/src/lib.rs
+++ b/binary_utils/src/lib.rs
@@ -2,19 +2,24 @@ use anyhow::Result;
 use chrono::{Local, SecondsFormat};
 use fs_err as fs;
 use logging::{debug_with_peers, exception};
+use logroller::{Compression, LogRollerBuilder, Rotation, RotationSize};
 use rayon::ThreadPoolBuilder;
 use std::{
     io::{self, IsTerminal},
     path::Path,
+    sync::OnceLock,
 };
+use tracing_appender::non_blocking::WorkerGuard;
 use tracing_subscriber::{
     filter::LevelFilter,
     fmt,
-    fmt::{format::Writer, time::FormatTime, writer::BoxMakeWriter},
+    fmt::{format::Writer, time::FormatTime},
     prelude::*,
     reload::{self, Handle},
     EnvFilter, Registry,
 };
+
+static LOG_GUARD: OnceLock<WorkerGuard> = OnceLock::new();
 
 #[derive(Clone)]
 pub struct TracingHandle(Handle<EnvFilter, Registry>);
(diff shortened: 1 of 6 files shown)


binary_utils/src/lib.rs
```diff
@@ -40,6 +41,59 @@ impl FormatTime for LocalTimer {
     }
 }
 
+struct RotatingWriter {
+    path: PathBuf,
+    max_size: u64,
```
Tumas: You could use [`bytesize`](https://crates.io/crates/bytesize) for `max_size` type to make it more readable as it is done for [database](https://github.com/grandinetech/grandine/blob/develop/database/src/lib.rs#L102) module

binary_utils/src/lib.rs
```diff
@@ -40,6 +41,59 @@ impl FormatTime for LocalTimer {
...
+        OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(&self.path)
+            .expect("Failed to open log file")
+    }
+
+    fn rotate_if_needed(&self) {
+        if let Ok(meta) = fs::metadata(&self.path) {
+            if meta.len() >= self.max_size {
```
Tumas: You should use [`logroller`](https://crates.io/crates/logroller) crate for this. 

Let's do this: 
- rotatation strategy: size-based on 1GB 
- enable compression for old logs
- file retention: keep max 5 entries of historical logs

sntntn: Thanks! Fantastic crate, it really simplifies the whole rotation logic

sntntn: The implementation is now based on logroller, with a size-based rotation (1 GB), gzip compression for old logs, and retention of up to 5 logs files (1 current + 4 compressed).

Everything has been tested and verified again by temporarily lowering the threshold to 1 KB and placing exception macro at frequent parts to ensure the rotation and compression behave as expected

Tumas (approved):

PR #475, opened 2025-11-12
zkVM Zisk Integration

jimmychu0807:
@ArtiomTr @povi 
Zisk integration is completed in this PR.
sha2 and bls12_381 is pointing to my own repository, so you may want to review and merge these two PRs:

- sha2: https://github.com/grandinetech/universal-precompiles/pull/5
- bls12_381: https://github.com/grandinetech/universal-precompiles/pull/6

and then update `Cargo.toml` pointing back to grandinetech repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,15 +96,14 @@ jobs:
   zkvm-test:
     strategy:
       matrix:
-        backend: [risc0, sp1, pico, ziren]
+        backend: [risc0, sp1, pico, ziren, zisk]
         include:
           # zkVM test - risc0 implementation
           - backend: risc0
-            # Refer to docs:
-            # https://dev.risczero.com/api/zkvm/install#using-rzup-in-ci
+            # Refer to docs: https://dev.risczero.com/api/zkvm/install#using-rzup-in-ci
             install: |
               curl -L https://risczero.com/install | bash
-              source /home/runner/.bashrc
+              echo "/home/runner/.risc0/bin" >> $GITHUB_PATH
               export PATH="/home/runner/.risc0/bin:$PATH"
               rzup install rust 1.88.0
               rzup install cpp 2024.1.5
@@ -117,7 +116,10 @@ jobs:
           - backend: sp1
             install: |
               curl -L https://sp1up.succinct.xyz | bash
-              /home/runner/.sp1/bin/sp1up -v 5.2.1
+              echo "/home/runner/.sp1/bin" >> $GITHUB_PATH
+              export PATH="/home/runner/.sp1/bin:$PATH"
+              sp1up -v 5.2.1
+              cargo prove --version
             execute: RUST_LOG=info cargo run -p zkvm_host --features sp1 --release -- --test "$TEST" execute
 
           # zkVM test - Brevis Pico implementation
@@ -126,7 +128,19 @@ jobs:
               rustup install nightly-2025-08-04
               rustup component add rust-src --toolchain nightly-2025-08-04
               cargo +nightly-2025-08-04 install --git https://github.com/brevis-network/pico pico-cli
+              cargo pico --version
             execute: RUST_LOG=info cargo +nightly-2025-08-04 run -p zkvm_host --features pico --release -- --test "$TEST" execute
+
+          # zkVM test - ZiskOS implementation
+          - backend: zisk
+            install: |
+              sudo DEBIAN_FRONTEND='noninteractive' apt-get update
+              sudo DEBIAN_FRONTEND='noninteractive' apt-get install -y xz-utils jq curl build-essential qemu-system libomp-dev libgmp-dev nlohmann-json3-dev protobuf-compiler uuid-dev libgrpc++-dev libsecp256k1-dev libsodium-dev libpqxx-dev nasm libopenmpi-dev openmpi-bin openmpi-common libclang-dev clang gcc-riscv64-unknown-elf
+              curl https://raw.githubusercontent.com/0xPolygonHermez/zisk/main/ziskup/install.sh | GH_RUNNER=true bash
+              echo "/home/runner/.config/.zisk/bin" >> $GITHUB_PATH
+              export PATH="/home/runner/.config/.zisk/bin:$PATH"
+              cargo-zisk -V
+            execute: RUST_LOG=info cargo run -p zkvm_host --features zisk --release -- --test "$TEST" execute
 
           # zkVM test - ZKM implementation
           - backend: ziren
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -293,7 +293,7 @@ bincode = '1'
 bindgen = '0.72'
 bit_field = '0.10'
 bitvec = '1'
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.1' }
+bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.2' }
 blst = { version = '0.3', features = ['portable'] }
 bonsai-sdk = '1.4'
 borsh = '1.5'
@@ -427,7 +427,7 @@ serde_with = '3'
 serde_yaml = '0.9'
 # The `asm` feature in `sha2` doesn't do anything on recent `x86_64` CPUs
 # because `sha2` defaults to using CPU intrinsics.
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.1', features = ['compress'] }
+sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.2', features = ['compress'] }
 slog = { version = '2', features = ['max_level_trace'] }
 slog-async = '2'
 slog-stdlog = '4'
@@ -608,7 +608,7 @@ lto = false
 # We had to fork `jsonrpc` because it does not allow nonstandard members.
 jsonrpc-core = { git = 'https://github.com/grandinetech/jsonrpc.git' }
 quick-protobuf = { git = 'https://github.com/sigp/quick-protobuf.git', rev = '681f413312404ab6e51f0b46f39b0075c6f4ebfd' }
-sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.1' }
+sha2 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'sha2-v0.10.9-up.2' }
 
 [patch.'https://github.com/zkcrypto/bls12_381.git']
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.1' }
+bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.2' }
--- a/zkvm/README.md
+++ b/zkvm/README.md
@@ -94,6 +94,30 @@
 
    Again, replace `<test_case>` with one of the four test cases mentioned above.
 
+## ZiskVM implementation
+
+1. **Execute without proving:**
+
+   Install Zisk with proving key. Refer to [the docs](https://0xpolygonhermez.github.io/zisk/getting_started/installation.html#installing-zisk) for steps.
+
+   Then run:
+
+   ```sh
+   cargo run -p zkvm_host --features zisk --release -- --test <test_case> execute
+   ```
+   Replace `<test_case>` with one of the four test cases mentioned above.
+
+2. **Prove using local prover:**
+
+   In addition to what's needed in step 1, also generate the program setup files. Refer to [the docs](https://0xpolygonhermez.github.io/zisk/getting_started/quickstart.html#prove) on steps for both generating and verifying the proof.
+
+   Then run:
+
+   ```sh
+   cargo run -p zkvm_host --features zisk --release -- --test <test_case> prove
+   ```
+   Again, replace `<test_case>` with one of the four test cases mentioned above.
+
 ## Adding new zkvm
 
 To add new zkvm, you need to:
--- a/zkvm/guest/pico/Cargo.lock
+++ b/zkvm/guest/pico/Cargo.lock
@@ -404,7 +404,7 @@ dependencies = [
  "rand_chacha 0.3.1",
  "serde",
  "serde_utils",
- "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.1)",
+ "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.2)",
  "ssz",
  "static_assertions",
  "typenum",
@@ -427,7 +427,7 @@ dependencies = [
 [[package]]
 name = "bls12_381"
 version = "0.8.0"
-source = "git+https://github.com/grandinetech/universal-precompiles.git?tag=bls12_381-6bb9695-up.1#673d417a78d34473f503972efb224b97b588d762"
+source = "git+https://github.com/grandinetech/universal-precompiles.git?tag=bls12_381-6bb9695-up.2#ae7f8574a370f4e9af93c9ec4aadb6d696f657d4"
 dependencies = [
  "bytemuck",
  "cfg-if",
@@ -440,6 +440,7 @@ dependencies = [
  "rand_core 0.6.4",
  "sp1-lib",
  "subtle",
+ "ziskos",
  "zkm-lib",
 ]
 
@@ -1704,7 +1705,7 @@ dependencies = [
  "ethereum-types",
  "generic-array",
  "hex-literal",
- "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.1)",
+ "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.2)",
 ]
 
 [[package]]
@@ -1733,7 +1734,7 @@ dependencies = [
  "rayon",
  "rc-box",
  "serde",
- "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.1)",
+ "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.2)",
  "shuffling",
  "ssz",
  "static_assertions",
@@ -2158,6 +2159,11 @@ checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 dependencies = [
  "spin",
 ]
+
+[[package]]
+name = "lib-c"
+version = "0.13.0"
+source = "git+https://github.com/0xPolygonHermez/zisk.git?tag=v0.13.0#ea1ed4c518992a170fc59ec19f1228eb4829a9e1"
 
 [[package]]
 name = "libc"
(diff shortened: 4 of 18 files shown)


.github/workflows/ci.yml
```diff
@@ -96,37 +96,56 @@ jobs:
...
           - backend: pico
             install: |
               rustup install nightly-2025-08-04
               rustup component add rust-src --toolchain nightly-2025-08-04
               cargo +nightly-2025-08-04 install --git https://github.com/brevis-network/pico pico-cli
-            execute: RUST_LOG=info cargo +nightly-2025-08-04 run -p zkvm_host --features pico --release -- --test "$TEST" execute
+              cargo pico --version
+            test_cmd: RUST_LOG=info cargo +nightly-2025-08-04 run -p zkvm_host --features pico --release -- --test "$TEST"
+            test_mode: execute
+
```
ArtiomTr, line 133: is there particular reason changing previous backends? It seems that they work fine currently?

.github/workflows/ci.yml
```diff
@@ -140,6 +159,7 @@ jobs:
       - name: Install system dependencies (Ubuntu)
         if: runner.os == 'Linux'
         run: |
+          sudo DEBIAN_FRONTEND='noninteractive' apt-get update
```
ArtiomTr: same as before, why changing pipeline that worked previously?

Cargo.toml
```diff
@@ -298,7 +298,11 @@ base64 = '0.22'
 bincode = '1'
 bit_field = '0.10'
 bitvec = '1'
-bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.1' }
+# bls12_381 = { git = 'https://github.com/grandinetech/universal-precompiles.git', tag = 'bls12_381-6bb9695-up.1' }
+# bls12_381 = { package = "sp1_bls12_381", path = "../zisk-bls12_381" }
+# bls12_381 = { path = "../grandine-universal-precompiles-bls12_381" }
+bls12_381 = { git = 'https://github.com/jimmychu0807/grandine-universal-precompiles.git', branch = 'jc/zkvm-zisk-bls12_381' }
+
```
ArtiomTr: putting here this comment just as a reminder to change it once precompiles are merged

zkvm/guest/risc0/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = 'zkvm_guest_risc0'
 edition = { workspace = true }
+version = "0.1.0"
```
ArtiomTr: why is it needed here?

zkvm/guest/risc0/guest/Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "risc0_grandine_state_transition"
+edition = "2024"
 version = "0.1.0"
-edition = "2021"
+authors = ["Grandine <info@grandine.io>"]
```
ArtiomTr: and here, why changing risc0?

zkvm/guest/sp1/Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = 'zkvm_guest_sp1'
-edition = '2021'
+edition = "2024"
+version = "0.1.0"
```
ArtiomTr: same here

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,309 @@
...
+            return Err(anyhow!(
+                "Failed to execute ziskemu command. Stderr:\n{output}"
+            ));
+        }
+
+        // Extract the execution cycle from the output.
+        let cycles: u64 = Self::extract_cycles(&output)?;
+
+        // Gather back the last set_output() from the VM guest output.
+        let state_root = Self::collect_result(&output)?;
```
ArtiomTr: looks like you can avoid hex serialization/deserialization by using output flag when executing zisk vm. You can look at how [ere](https://github.com/eth-act/ere/blob/a75f8f4832942b32fc1314bb3b2ce9b5e3e9d306/crates/zkvm/zisk/src/zkvm/sdk.rs#L217-L218) does this.

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,309 @@
...
+        //   cargo-zisk rom-setup -e target/riscv64ima-zisk-zkvm-elf/release/zkvm_guest_zisk
+        // 2. Generate proof
+        //   LIB_EXT=$([[ "$(uname)" == "Darwin" ]] && echo "dylib" || echo "so")
+        //   cargo-zisk prove -e target/riscv64ima-zisk-zkvm-elf/release/zkvm_guest_zisk -i build/input.bin -o ./ -a
+        let zisk_guest_dir = Self::get_guest_dir();
+        let input_path =
+            Self::generate_input_file(config, state_ssz, block_ssz, cache_ssz, phase_bytes)?;
+        let elf_path = zisk_guest_dir.join(ELF_PATH_SUFFIX);
+
+        println!("🛠️ Running program setup `cargo-zisk rom-setup`...");
```
ArtiomTr: please don't use emojis -- all backend outputs should be +- similar

zkvm/guest/zisk/src/main.rs
```diff
@@ -0,0 +1,93 @@
...
+    input: &[u8],
+) -> Result<(Config, SignedBeaconBlock<P>, BeaconState<P>, PubkeyCache)> {
+    // Deserialize the input using bincode
+    let VMGuestInput {
+        config,
+        state_ssz,
+        block_ssz,
+        cache_ssz,
+        phase_bytes,
+    } = bincode::deserialize(input).unwrap();
```
ArtiomTr: don't use double encoding here -- you're currently serializing state&block&cache into SSZ, and then serialize SSZ bytes into bincode. Instead, you can pass those 3 directly as bytes into zkvm, the same as with other backends.

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,309 @@
...
+        };
+
+        exit_status.success()
+    }
+
+    fn save(&self, path: impl AsRef<Path>) -> Result<()> {
+        let zisk_guest_dir = Vm::get_guest_dir();
+        let proofs_path = Vm::get_guest_dir()
+            .join("proofs")
+            .join("vadcop_final_proof.bin");
```
ArtiomTr: just genuine question -- what is "vadcop" here?

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let output = run_cmd(
+            Command::new("cp")
+                .arg(&proofs_path)  // src
+                .arg(path.as_ref())  // destination
+                .current_dir(&zisk_guest_dir),
+        )?;
```
ArtiomTr: I believe doing `fs::copy_file` here is both more portable and quicker

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,309 @@
...
+        // Gather back the last set_output() from the VM guest output.
+        let state_root = Self::collect_result(&output)?;
+        Ok((state_root, Proof))
+    }
+}
+
+// This function runs the command, streaming the output to screen immediately while capuring the
+// output and return it in a buffer.
+fn run_cmd(cmd: &mut Command) -> Result<(ExitStatus, Vec<u8>)> {
+    let mut child = cmd.stdout(Stdio::piped()).spawn()?;
```
ArtiomTr, line 276: I haven't tested this, but my intuition was, that you can do `Stdio::inherit()`, and then, instead of plumbing output manually, you can just do `.output()`, and everything will be returned & streamed into your stdout automatically?

ArtiomTr (changes requested):

.github/workflows/ci.yml
jimmychu0807: For this, this is optimization. I think it is more elegant and consistent this way. I added `test_mode` so we can extend it to run in **prove** mode, which I did here (it times out at the end): 
https://github.com/jimmychu0807/grandine/actions/runs/19923544736

.github/workflows/ci.yml
jimmychu0807: let me remove this.

zkvm/guest/risc0/Cargo.toml
jimmychu0807: For consistency sake. If you prefer not having it, I can revert it.

zkvm/guest/risc0/guest/Cargo.toml
jimmychu0807: For consistency sake. If you prefer not having it, I can revert it.

zkvm/guest/sp1/Cargo.toml
jimmychu0807: For consistency sake. If you prefer not having it, I can revert it.

zkvm/host/src/backend/zisk.rs
jimmychu0807: I see. They output the result to a file and read that file.

zkvm/host/src/backend/zisk.rs
jimmychu0807: I am not sure too. It is just how they name the output/final proof name, by default. Refer:
https://0xpolygonhermez.github.io/zisk/getting_started/quickstart.html#verify-proof

zkvm/guest/risc0/Cargo.toml
ArtiomTr: yeah, better to have only zisk-related changes in this PR. If you want to fix/make better other backends, lets do this in separate PR

.github/workflows/ci.yml
ArtiomTr: I believe we don't need prove mode now, as it looks like zkvms currently are not capable to prove on CPU in reasonable time. I believe GPUs are not available in github runners (and probably will never be available), so we can safely remove proving mode from pipeline.

zkvm/host/src/backend/zisk.rs
jimmychu0807: I ask AI about this. It seems there is no way to do it. Check my 2nd question below in the conversation:
https://www.perplexity.ai/search/is-it-possible-to-use-stdio-in-lmdW6nE2SaGG9Mo1oUSClw#1

.github/workflows/ci.yml
jimmychu0807: `test_mode` removed.

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,311 @@
...
+        let (exit_status, output_bytes) = prove_output;
+        let output = String::from_utf8_lossy(&output_bytes);
+
+        if !exit_status.success() {
+            return Err(anyhow!("Failed to generate proof. Stderr:\n{output}"));
+        }
+
+        // Gather back the last set_output() from the VM guest output.
+        // Note: maybe able to read from the output dir the public output.
+        let state_root = Self::collect_result(&output)?;
```
jimmychu0807: I want to keep this line here. Because I have never able to execute to this point actually (**prove** mode). So keeping this line let future dev know they could either read from the output dir (I'm not sure what's the output file name), or collect the result from the screen output.

zkvm/guest/zisk/src/main.rs
jimmychu0807: From what I found, `ziskemu` only takes input in a file format, not bytes directly.

- ref 1: [Zisk docs](https://0xpolygonhermez.github.io/zisk/getting_started/quickstart.html#execute)
- ref 2: [ere also writes the input bytes to a temp file](https://github.com/eth-act/ere/blob/a75f8f4832942b32fc1314bb3b2ce9b5e3e9d306/crates/zkvm/zisk/src/zkvm/sdk.rs#L182-L183)

So the issue becomes, I either not using **bincode** and add byte separators and then concatenate the 5 inputs myself, or using bincode and let the lib handles this for me.

For ref, input is serialized here:
https://github.com/grandinetech/grandine/pull/475/changes#r2625605370

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,311 @@
...
+    fn execute(
+        &self,
+        config: ConfigKind,
+        state_ssz: Vec<u8>,
+        block_ssz: Vec<u8>,
+        cache_ssz: Vec<u8>,
+        phase_bytes: Vec<u8>,
+    ) -> Result<(Vec<u8>, Self::Report)> {
+        let input_path =
+            Vm::generate_input_file(config, state_ssz, block_ssz, cache_ssz, phase_bytes)?;
```
jimmychu0807, line 160: input is serialized here.

zkvm/guest/zisk/src/main.rs
ArtiomTr: > From what I found, `ziskemu` only takes input in a file format, not bytes directly.

Yeah, but you still can write bytes directly into file, so in that sense, it accepts bytes as input.

> So the issue becomes, I either not using **bincode** and add byte separators and then concatenate the 5 inputs myself, or using bincode and let the lib handles this for me.

I just see it as with every other backend - you just read directly from byte array. As far as I remember, when we tried using bincode instead of ssz, we saw pretty large performance decrease. But you can try both approaches and compare, which one performs better:

* Using bincode (your current implementation).
* Doing as every other vm - reading 4 `usize`'s from byte stream, then reading 4 slices with appropriate lengths, something like so:
  ```rust
  let mut buffer = [0u8; size_of::<usize>()];
  input.read_exact(&mut buffer)?;
  let state_ssz_len = usize::from_be_bytes(&v);

  // same for block_ssz_len, cache_ssz_len and phase_bytes_len

  let mut state_ssz = vec![0u8; state_ssz_len];
  input.read_exact(&mut state_ssz)?;

  // same for block_ssz, cache_ssz and phase_bytes
  ```

zkvm/host/src/backend/zisk.rs
ArtiomTr: I believe you can just comment out whole guest program, keep empty `main`, and then run prove mode locally, so that you can find where file is located.

It is better to check proving mode before merging, because after preparing all zkvms, we want to conduct some kind of report. So we will likely hand out code to zisk team and ask them to run proving on their rig.

zkvm/host/src/backend/zisk.rs
ArtiomTr: yeah, tried this locally, and that doesn't work 😢

ArtiomTr (changes requested):
Looks good, few changes needed. Also, looks like [pipeline currently fails](https://github.com/jimmychu0807/grandine/actions/runs/20290830767/job/58274779111)? This needs to be fixed before merging

zkvm/guest/zisk/src/main.rs
jimmychu0807: updated this section of code to plainly write bytes to a file and updated the guest code to read bytes from the file.

This is indeed faster than using bincode as it is simpler.

zkvm/host/src/backend/zisk.rs
jimmychu0807: Great suggestion. I am able to create the proof file. I realize Zisk proof mode doesn't output the result, so I have to run `self.execute()` first to get the result and the proof output. 

Have also updated `Proof::verify()` and `Proof::save()` methods accordingly with the output path.

zkvm/guest/zisk/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+name = "zkvm_guest_zisk"
+edition = "2024"
+version = "0.1.0"
+authors = ["Grandine <info@grandine.io>"]
+
+[workspace]
+
+[dependencies]
+anyhow = { version = '1', features = ['backtrace'] }
+bincode = '1'
```
ArtiomTr: I think this no longer needed?

zkvm/guest/zisk/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+bls = { path = '../../../bls', features = ['zkcrypto'], default-features = false }
+byteorder = '1'
+enum-iterator = '2'
+pubkey_cache = { path = "../../../pubkey_cache" }
+snap = '1'
```
ArtiomTr: this probably also not needed?

zkvm/README.md
```diff
@@ -94,6 +94,31 @@
...
+   Replace `<test_case>` with one of the four test cases mentioned above.
+
+2. **Prove using local prover:**
+
+   In addition to what's needed in step 1, also generate the program setup files. Refer to [the docs](https://0xpolygonhermez.github.io/zisk/getting_started/quickstart.html#prove) on steps for both generating and verifying the proof.
+
+   Then run:
+
+   ```sh
+   cargo -p zkvm_host --features zisk --release -- --test <test_case> prove
```
ArtiomTr: `run` command is missing here, probably should be:

```suggestion
   cargo run -p zkvm_host --features zisk --release -- --test <test_case> prove
```

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,294 @@
...
+        };
+
+        exit_status.success()
+    }
+
+    fn save(&self, path: impl AsRef<Path>) -> Result<()> {
+        let proof_path = Vm::get_artifacts_dir().join(PROOF_PATH_SUFFIX);
+
+        let _ = fs::copy(&proof_path, path.as_ref())
+            .map_err(|e| anyhow!("Copy file failed. stderr:\n{}", e.to_string()))?;
```
ArtiomTr: You can make this error a bit nicer, by providing a bit more context:
```suggestion
        fs::copy(&proof_path, path.as_ref()).context(format!(
            "Failed to copy file from {proof_path:?} to {:?}",
            path.as_ref()
        ))?;
```

Mentioning this, because I've got this error:
```
Error: Copy file failed. stderr:
No such file or directory (os error 2)
```
As you can see, it's not very helpful :). With suggested change, it looks like:
```
Error: Failed to copy file from "/home/necolt/workspace/grandine/zkvm/guest/zisk/artifacts/vadcop_final_proof.compressed.bin" to "/home/necolt/workspace/grandine/zkvm/host/proof.bin"

Caused by:
    No such file or directory (os error 2)
```
Which is much nicer in my opinion

zkvm/host/src/backend/zisk.rs
```diff
@@ -0,0 +1,294 @@
...
+        file.write_all(&phase_bytes.len().to_be_bytes())?;
+        file.write_all(&phase_bytes)?;
+
+        file.sync_all()?;
+
+        Ok(input_path)
+    }
+
+    fn get_guest_dir() -> PathBuf {
+        Path::new(env!("CARGO_MANIFEST_DIR")).join("../guest/zisk")
```
ArtiomTr: to make path a bit nicer, you can do:

```suggestion
        Path::new(env!("CARGO_MANIFEST_DIR")).parent().expect("project cannot be at root directory").join("guest/zisk")
```

this way paths will look better when trying to print them:

instead of:
```
/<project-path>/grandine/zkvm/host/../guest/zisk
```
you'll see:
```
/<project-path>/grandine/zkvm/guest/zisk
```

ArtiomTr (changes requested):
The code looks good, however, when I try to prove locally, I get this error:
```
Error: Failed to copy file from "/home/necolt/workspace/grandine/zkvm/guest/zisk/artifacts/vadcop_final_proof.compressed.bin" to "/home/necolt/workspace/grandine/zkvm/host/proof.bin"

Caused by:
    No such file or directory (os error 2)
```
And indeed, I don't have such file:
```
/zkvm/guest/zisk/artifacts/vadcop_final_proof.compressed.bin
```
Contents of whole `artifacts` directory:
```
input.bin
result.json
vadcop_final_proof.bin
```
Maybe zisk won't generate `.compressed` proof by default?

ArtiomTr (approved):
Looks good 👍

zkvm/guest/pico/Cargo.lock
```diff
@@ -404,7 +404,7 @@ dependencies = [
  "rand_chacha 0.3.1",
  "serde",
  "serde_utils",
- "sha2 0.10.9 (git+https://github.com/grandinetech/universal-precompiles.git?tag=sha2-v0.10.9-up.1)",
+ "sha2 0.10.9 (git+https://github.com/jimmychu0807/grandine-universal-precompiles.git?branch=jc%2Fzkvm-zisk-sha2)",
```
povi: Please don't include cryptography (or other) dependencies from a personal repo, thanks.

jimmychu0807: @povi Nice catch! I wasn't aware of that.

Now I have rebuilt the guest code of pico, sp1, and ziren and have the corresponding **Cargo.lock** updated.

povi (approved):

PR #476, opened 2025-11-12
Implement grandine-nethermind integration

ArtiomTr:
Before merging:

* [x] Double-check execution request (de)serialization logic
* [x] Fix CI pipeline
* [x] Refactor logging from nethermind plugin side
  * [x] Do not emit warning on each function call (was useful for debugging, but now too much junk in logs)
  * [x] Throw meaningful errors with proper messages, not "unexpected error occurred"
  * [x] Return meaningful errors from engine api - now it is just error code 1.
  * [x] Pass validation error in PayloadStatusV{1,2,3}
* [x] Run grandine cleanup from nethermind plugin
* [x] Don't panic during conversions in rust
* [x] Use array pooling when decoding withdrawals/etc.
* [x] Move from generic `Exception` type to more specific


bindings/csharp/Grandine.NethermindPlugin/GrandineUtils.cs
```diff
@@ -0,0 +1,30 @@
...
+    public unsafe static CH256 convertHash256(Hash256 hash) {
+        var result = new CH256();
+
+        if (hash == null) {
+            return result;
+        }
+
+        if (hash.Bytes.Length != 32)
+        {
+            throw new Exception("Hash256 must be exactly 32 bytes long");
```
LukaszRozmej: This is redundant, `Hash256` will always have 32 bytes

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
```diff
@@ -0,0 +1,158 @@
...
+    public string Name => "Grandine plugin";
+    public string Description => "Nethermind plugin to enable embedded grandine CL client";
+    public string Author => "Grandine team";
+
+    private ILogger _logger;
+
+    #pragma warning disable CS8618 
+    private INethermindApi _api;
+
+    public bool Enabled => grandineConfig.Enabled ?? throw new ArgumentNullException(nameof(grandineConfig.Enabled));
```
LukaszRozmej: This looks a bit weird for me to have `Enabled` be nullable.

This will cause node not to run unless it is explicitly configured. Normally we just default to `false`.

This is fine if this is your intention, but could be confusing for user. Maybe at least add a custom message for the user explaining what is going on?

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
```diff
@@ -0,0 +1,158 @@
...
+        Type configType = grandineConfig.GetType();
+
+        var rpcConfig = nethermindApi.Config<IJsonRpcConfig>();
+
+        var configInterface = configType.GetInterface("IGrandineConfig");
+
+        if (configInterface == null)
+        {
+            throw new Exception("Unexpected exception occurred - IGrandineConfig interface not found");
+        }
```
LukaszRozmej: This is weird, if you loaded your plugin, you do have the types loaded correctly - so seems not needed.
```suggestion
        var configInterface = typeof(IGrandineConfig);
```

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
```diff
@@ -0,0 +1,158 @@
...
+                var client = new GrandineClient(api.getAdapter());
+                client.Run(_arguments.ToArray());
+            }
+            catch (Exception e)
+            {
+                _logger.Error($"Failed to start grandine: {e}");
+            }
+        });
+        task.Start();
+        return task;
```
LukaszRozmej:
```suggestion
        return Task.Run(() =>
        {
            try
            {
                var client = new GrandineClient(api.getAdapter());
                client.Run(_arguments.ToArray());
            }
            catch (Exception e)
            {
                _logger.Error($"Failed to start grandine: {e}");
            }
        });
```

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+                Withdrawals = withdrawals.ToArray(),
+                BlobGasUsed = payload.blob_gas_used,
+                ExcessBlobGas = payload.excess_blob_gas,
+                ParentBeaconBlockRoot = parentBeaconBlockRootConverted,
+                ExecutionRequests = executionRequestsConverted.ToArray(),
+            }, versionedHashesConverted.ToArray(), parentBeaconBlockRootConverted, executionRequestsConverted.ToArray()).Result;
+
+            if (payloadStatus.Result != Result.Success)
+            {
+                throw new Exception("unexpected failure");
```
LukaszRozmej: Generally one should avoid throwing base `Exception` type in c#: https://stackoverflow.com/a/22453825

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+
+            var executionRequestsConverted = new List<byte[]>();
+            for (var i = 0; i < (int)executionRequests.requests_len; ++i)
+            {
+                executionRequestsConverted.Add(new ReadOnlySpan<byte>(executionRequests.requests[i].bytes, (int)executionRequests.requests[i].bytes_len).ToArray());
+            }
+
+            var payloadStatus = _engineRpc.engine_newPayloadV4(new Nethermind.Merge.Plugin.Data.ExecutionPayloadV3
+            {
+                ParentHash = new Hash256(new ReadOnlySpan<byte>(payload.parent_hash, 32).ToArray()),
```
LukaszRozmej: You shouldn't need to allocate array for the hash, will save a bit on allocations.
Applies everywhere here

```suggestion
                ParentHash = new Hash256(new ReadOnlySpan<byte>(payload.parent_hash, 32)),            
```

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+                StateRoot = new Hash256(new ReadOnlySpan<byte>(payload.state_root, 32).ToArray()),
+                ReceiptsRoot = new Hash256(new ReadOnlySpan<byte>(payload.receipts_root, 32).ToArray()),
+                LogsBloom = new Bloom(new ReadOnlySpan<byte>(payload.logs_bloom, (int)payload.logs_bloom_len).ToArray()),
+                PrevRandao = new Hash256(new ReadOnlySpan<byte>(payload.prev_randao, 32).ToArray()),
+                BlockNumber = (long)payload.block_number,
+                GasLimit = (long)payload.gas_limit,
+                GasUsed = (long)payload.gas_used,
+                Timestamp = payload.timestamp,
+                ExtraData = new ReadOnlySpan<byte>(payload.extra_data, (int)payload.extra_data_len).ToArray(),
+                BaseFeePerGas = new UInt256(new ReadOnlySpan<byte>(payload.base_fee_per_gas, 32).ToArray(), true),
```
LukaszRozmej: Same for UInts256:
```suggestion
                BaseFeePerGas = new UInt256(new ReadOnlySpan<byte>(payload.base_fee_per_gas, 32), true),
```

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+                versionedHashesConverted.Add(new ReadOnlySpan<byte>(versionedHashes[i], 32).ToArray());
+            }
+
+            var parentBeaconBlockRootConverted = new Hash256(new ReadOnlySpan<byte>(parentBeaconBlockRoot, 32).ToArray());
+
+            var executionRequestsConverted = new List<byte[]>();
+            for (var i = 0; i < (int)executionRequests.requests_len; ++i)
+            {
+                executionRequestsConverted.Add(new ReadOnlySpan<byte>(executionRequests.requests[i].bytes, (int)executionRequests.requests[i].bytes_len).ToArray());
+            }
```
LukaszRozmej: You could allocate the array upfront as you know the length, saves you double allocation later:
```suggestion
            int length = (int)executionRequests.requests_len;
            var executionRequestsConverted = byte[][(length];
            for (var i = 0; i < (int)length ; ++i)
            {
                executionRequestsConverted[i] = new ReadOnlySpan<byte>(executionRequests.requests[i].bytes, (int)executionRequests.requests[i].bytes_len).ToArray();
            }
```

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+                Timestamp = payload.timestamp,
+                ExtraData = new ReadOnlySpan<byte>(payload.extra_data, (int)payload.extra_data_len).ToArray(),
+                BaseFeePerGas = new UInt256(new ReadOnlySpan<byte>(payload.base_fee_per_gas, 32).ToArray(), true),
+                BlockHash = new Hash256(new ReadOnlySpan<byte>(payload.block_hash, 32).ToArray()),
+                Transactions = transactions.ToArray(),
+                Withdrawals = withdrawals.ToArray(),
+                BlobGasUsed = payload.blob_gas_used,
+                ExcessBlobGas = payload.excess_blob_gas,
+                ParentBeaconBlockRoot = parentBeaconBlockRootConverted,
+                ExecutionRequests = executionRequestsConverted.ToArray(),
```
LukaszRozmej:
```suggestion
                ExecutionRequests = executionRequestsConverted,
```

LukaszRozmej (approved):
Super cool, left few comments on potential improvements

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        _api = nethermindApi;
+        _logger = nethermindApi.LogManager.GetClassLogger();
+        _logger.Warn("Initializing grandine plugin...");
+
+        Type configType = grandineConfig.GetType();
+
+        var rpcConfig = nethermindApi.Config<IJsonRpcConfig>();
+
+        var configInterface = configType.GetInterface("IGrandineConfig");
```
LukaszRozmej:
```suggestion
        var configInterface = configType.GetInterface(nameof(IGrandineConfig));
```

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
ArtiomTr: I was trying to fix C# warnings, because when Enabled is defined without `?` in GrandineConfig, then C# yells that it must be initialized in constructor. But probably I should just disable warning in this place

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
ArtiomTr: Same story as with `Enabled` option, wanted to fix all warnings, but probably redundant here

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
ArtiomTr: Good to know, thank you :pray:

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
```diff
@@ -0,0 +1,1814 @@
...
+        };
+    }
+
+    unsafe CResult<CPayloadStatusV1> EngineNewPayloadV4(CExecutionPayloadV3 payload, byte** versionedHashes, ulong versionedHashesLen, byte* parentBeaconBlockRoot, CExecutionRequests executionRequests)
+    {
+        _logger.Warn("================================================================= engine_newPayloadV4 =================================================================");
+
+        try
+        {
+            var transactions = new List<byte[]>();
```
LukaszRozmej: To really make it super light when arrays are actually needed, you could use some ArrayPool for that!

Just note that built-in array pool works only on power of 2 lenght's and thus is designed for Spans, but here you need exact lenghts which is a bit more problematic. So you would need something like: https://stackoverflow.com/q/77445864

bindings/csharp/Grandine.NethermindPlugin/GrandinePlugin.cs
LukaszRozmej: you can just use field/property initializer

bindings/csharp/Grandine.NethermindPlugin/GrandineEngineApi.cs
LukaszRozmej: Maybe better to create your own array pool with exact lengths?
Best thing would be if we migrated Arrays to `Memory<>` or something, but that is big refactor

sauliusgrigaitis: > Maybe better to create your own array pool with exact lengths? Best thing would be if we migrated Arrays to `Memory<>` or something, but that is big refactor

At this point we just want to ship this thing. I believe migrating Arrays to `Memory<>` is not something that will happen very soon? 

Do you have any idea how to easily create our own array pool? It doesn't look like something we can quickly do ourself without deeper C# knowledge.

LukaszRozmej: Super easy ;)
But would be best used where there are actually count's that are very repetetive (like withdrawals), and not arbitrary like transactions

```
using System;
using System.Collections.Concurrent;

public class ExactSizeArrayPool<T>
{
    private readonly ConcurrentDictionary<int, ConcurrentBag<T[]>> _pools;
    private readonly int _maxArraysPerSize;

    public ExactSizeArrayPool(int maxArraysPerSize = 10)
    {
        _pools = new ConcurrentDictionary<int, ConcurrentBag<T[]>>();
        _maxArraysPerSize = maxArraysPerSize;
    }

    public T[] Rent(int exactSize)
    {
        if (exactSize <= 0)
            throw new ArgumentException("Size must be positive", nameof(exactSize));

        // GetOrAdd is thread-safe and efficient
        var pool = _pools.GetOrAdd(exactSize, _ => new ConcurrentBag<T[]>());

        // Try to get an existing array
        if (pool.TryTake(out var array))
        {
            Array.Clear(array, 0, array.Length);
            return array;
        }

        // Create new array if none available
        return new T[exactSize];
    }

    public void Return(T[] array)
    {
        if (array == null)
            throw new ArgumentNullException(nameof(array));

        int size = array.Length;
        var pool = _pools.GetOrAdd(size, _ => new ConcurrentBag<T[]>());

        // Only return if we haven't exceeded the limit
        if (pool.Count < _maxArraysPerSize)
        {
            pool.Add(array);
        }
    }

    public void Clear()
    {
        _pools.Clear();
    }
}
```

bindings/c/src/generic.rs
```diff
@@ -0,0 +1,363 @@
...
+            is_something: false,
+            value: Default::default(),
+        }
+    }
+}
+
+impl<T: Clone> Clone for COption<T> {
+    fn clone(&self) -> Self {
+        Self {
+            is_something: self.is_something.clone(),
```
povi: No need to clone boolean

bindings/c/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+bls = { workspace = true }
+clap = { workspace = true }
+eth1_api = { workspace = true, features = ["embed"] }
+ethereum-types = { workspace = true }
+execution_engine = { workspace = true }
+generic-array = { workspace = true }
+hex = { workspace = true }
+kzg_utils = { workspace = true }
+runtime = { workspace = true, features = [
+    "network-goerli",
```
povi: Remove legacy networks: goerli, holesky, shadow

eth1_api/src/eth1_api/embed_api.rs
```diff
@@ -0,0 +1,737 @@
...
+    ForkChoiceUpdatedResponse, PayloadAttributes, PayloadAttributesV1, PayloadAttributesV2,
+    PayloadAttributesV3, PayloadId, PayloadStatusV1,
+};
+use futures::channel::mpsc::UnboundedSender;
+use prometheus_metrics::Metrics;
+use reqwest::Client;
+use serde::Deserialize;
+use ssz::H256;
+use static_assertions::const_assert_eq;
+use std::{
```
povi, line 19: Put `std` imports at the top

Tumas (approved):

povi (approved):

bindings/csharp/Grandine.NethermindPlugin/GrandineUtils.cs
```diff
@@ -0,0 +1,247 @@
...
+using System;
+using System.Runtime.InteropServices;
+using Grandine.Native;
+using Nethermind.Consensus.Producers;
+using Nethermind.Core;
+using Nethermind.Merge.Plugin.Data;
+
+public static class GrandineUtils
+{
+    private static readonly ExactSizeArrayPool<Withdrawal> WithdrawalPool = new ();
```
LukaszRozmej: You probably don't need more than 1 array per size? Unless multiple Payloads are created at the same time.

```suggestion
    private static readonly ExactSizeArrayPool<Withdrawal> WithdrawalPool = new(1);
```

LukaszRozmej (approved):

bindings/csharp/Grandine.NethermindPlugin/GrandineUtils.cs
ArtiomTr: That's true, missed that part, thank you for the suggestion

flcl42 (approved):

PR #483, opened 2025-11-17
feat: Add topic changes for gloas

### LINKED ISSUE #401: Add Gloas gossipsub topics
spec: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#the-gossip-domain-gossipsub

Subhasish-Behera:
feat: Add remaining gossipsub handling flow
    
    P2P Layer (p2p/src/block_sync_service.rs, p2p/src/network.rs)
      - Added envelope deduplication tracking per (block_root, builder_index)
      - Implemented gossip reception handlers for ExecutionPayload and PayloadAttestation
    
    Fork Choice Store (fork_choice_store/src/store.rs, fork_choice_store/src/misc.rs,
      fork_choice_store/src/error.rs)
      - Implemented validate_execution_payload_envelope() with validation pipeline
      - Added accepted_execution_payload_envelopes tracking
      - Added delay mechanisms (DelayUntilBeaconBlock, DelayUntilSlot)
    
    Fork Choice Controller (fork_choice_control/src/controller.rs, fork_choice_control/src/mutator.rs,
      fork_choice_control/src/tasks.rs, fork_choice_control/src/thread_pool.rs)
      - Implemented ExecutionPayloadEnvelopeTask worker validation
      - Added mutator handlers for Accept/Ignore/Delay actions
      - Added retry queue for delayed envelopes
    
    Events & Metrics (fork_choice_control/src/events.rs, prometheus_metrics/src/metrics.rs)
      - Added ExecutionPayloadEnvelopeEvent with execution layer context
    
    - Fixed P2P deduplication to use (block_root, builder_index) tuple


fork_choice_control/src/events.rs
```diff
@@ -614,6 +642,30 @@ impl<P: Preset> DataColumnSidecarEvent<P> {
     }
 }
 
+#[derive(Clone, Copy, Debug, Serialize)]
+pub struct ExecutionPayloadEnvelopeEvent {
```
hangleang: you should look at [this PR](https://github.com/ethereum/beacon-APIs/pull/552/files#diff-3cc6eaa800c1a4bd6adcd78f1722b466f7a24048a66e320c8f92872fbcd9eefb) for the event data structure

fork_choice_store/src/error.rs
```diff
@@ -187,4 +200,6 @@ pub enum Error<P: Preset> {
     },
 }
 
-assert_eq_size!(Error<Mainnet>, [usize; 4]);
+// Size increased due to new ExecutionPayloadEnvelope error variants
+// TODO: Update size after verifying actual Error<Mainnet> size
+// assert_eq_size!(Error<Mainnet>, [usize; 4]);
```
hangleang: @povi can you verify this check sir?

fork_choice_store/src/misc.rs
```diff
@@ -815,6 +815,72 @@ pub enum PartialAttestationAction {
...
+    pub const fn gossip_id_ref(&self) -> Option<&GossipId> {
+        match self {
+            Self::Gossip(gossip_id) => Some(gossip_id),
+            Self::Requested(_) => None,
+            Self::Own => None,
+        }
+    }
+
+    #[must_use]
+    pub const fn is_from_block(&self) -> bool {
```
hangleang: envelope seperate from block

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+        &self,
+        envelope: Arc<SignedExecutionPayloadEnvelope<P>>,
+        beacon_block_seen: bool,
+        origin: &ExecutionPayloadEnvelopeOrigin,
+    ) -> Result<ExecutionPayloadEnvelopeAction<P>> {
+        let slot = envelope.message.slot;
+        let beacon_block_root = envelope.message.beacon_block_root;
+        let builder_index = envelope.message.builder_index;
+
+        if !origin.is_from_block() {
```
hangleang: as comment above

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+            // [IGNORE] The envelope is from a slot greater than or equal to the latest finalized slot
+            // Spec: envelope.slot >= compute_start_slot_at_epoch(store.finalized_checkpoint.epoch)
+            if slot < self.finalized_slot() {
+                return Ok(ExecutionPayloadEnvelopeAction::Ignore(false));
+            }
+        }
+
+        // [IGNORE] The envelope's beacon_block_root has been seen (via gossip or non-gossip sources)
+        // (a client MAY queue envelope for processing once the block is retrieved)
+        if !beacon_block_seen {
```
hangleang: what is different from the `Store::chain_link` check?

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+        let Some(chain_link) = self.chain_link(beacon_block_root) else {
+            // Block not in store yet, delay until it arrives
+            return Ok(ExecutionPayloadEnvelopeAction::DelayUntilBeaconBlock(
+                envelope,
+                beacon_block_root,
+            ));
+        };
+
+        let block = &chain_link.block;
+        let state = chain_link.state(self);
```
hangleang: doing so would require load the beacon state, you should just get from `ChainLInk::state` and delay until the state available like https://github.com/grandinetech/grandine/blob/bc5c0dbaa1a535f851739e377e1c52f1ef41322b/fork_choice_store/src/store.rs#L1888-L1893

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+        );
+
+        // [REJECT] The builder_index must be a valid and active validator
+        let validator = state
+            .validators()
+            .get(builder_index)
+            .map_err(|_| Error::<P>::ValidatorNotActive { builder_index })?;
+
+        ensure!(
+            predicates::is_active_validator(validator, accessors::get_current_epoch(&state)),
```
hangleang: also make sure the validator is builder

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+                expected: block.message().slot(),
+                actual: slot,
+            },
+        );
+
+        // [REJECT] The builder_index must be a valid and active validator
+        let validator = state
+            .validators()
+            .get(builder_index)
+            .map_err(|_| Error::<P>::ValidatorNotActive { builder_index })?;
```
hangleang, line 2625: that is mean the validator is not exist based on the state

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+            predicates::is_active_validator(validator, accessors::get_current_epoch(&state)),
+            Error::<P>::ValidatorNotActive { builder_index },
+        );
+
+        // [REJECT] The builder signature envelope.signature is valid
+        // ExecutionPayloadEnvelope is Gloas-only, so state must be Gloas
+        let BeaconState::Gloas(gloas_state) = state.as_ref() else {
+            bail!("ExecutionPayloadEnvelope validation requires Gloas state");
+        };
+        transition_functions::gloas::execution_payload_processing::verify_execution_payload_envelope_signature(
```
hangleang: just call the verify function https://github.com/grandinetech/grandine/blob/bc5c0dbaa1a535f851739e377e1c52f1ef41322b/fork_choice_store/src/store.rs#L1896-L1906

fork_choice_store/src/store.rs
```diff
@@ -2436,6 +2439,118 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+            &self.chain_config,
+            &self.pubkey_cache,
+            gloas_state,
+            &envelope,
+            SingleVerifier,
+        )?;
+
+        // [REJECT] Get the bid from the block
+        // ExecutionPayloadEnvelope is Gloas-only, so block must be Gloas
+        let SignedBeaconBlock::Gloas(gloas_block) = block.as_ref() else {
```
hangleang: note that the payload bid can be obtain from the state as well, these is from the spec:

> Let block be the block with envelope.beacon_block_root. Let bid alias block.body.signed_execution_payload_bid.message (notice that this can be obtained from the state.latest_execution_payload_bid)

p2p/src/network.rs
```diff
@@ -2348,6 +2348,26 @@ impl<P: Preset> Network<P> {
...
+                P2pToSync::GossipExecutionPayload(
+                    execution_payload_envelope,
+                    source,
+                    GossipId { source, message_id },
+                )
+                .send(&self.channels.p2p_to_sync_tx);
+            }
+            PubsubMessage::PayloadAttestationMessage(payload_attestation) => {
+                debug!("received payload attestation message as gossip from {source}");
+                P2pToSync::GossipPayloadAttestation(
```
hangleang: why forward to sync manager?

p2p/src/block_sync_service.rs
```diff
@@ -654,6 +657,54 @@ impl<P: Preset> BlockSyncService<P> {
...
+                                    self.config.as_ref(),
+                                    &self.controller.head_state().value(),
+                                    payload_slot,
+                                );
+                                let slot_start_time = SystemTime::UNIX_EPOCH + Duration::from_secs(slot_start);
+                                let three_fourths_slot = slot_start_time + Duration::from_millis(
+                                    3 * self.config.seconds_per_slot.get() * 1000 / 4
+                                );
+
+                                if SystemTime::now() >= three_fourths_slot {
```
hangleang: you don't need to restrict the time of arrival as it can soon or later which we handle in the validation

fork_choice_store/src/error.rs
Subhasish-Behera: @hangleang  need to change it to 9 right . got the error ```202 | assert_eq_size!(Error<Mainnet>, [usize; 8]);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: source type: `error::Error<types::preset::Mainnet>` (576 bits)
    = note: target type: `[usize; 8]` (512 bits)```  after removing MissingExecutionPayloadBid as its not used

fork_choice_store/src/misc.rs
Subhasish-Behera: true. missed it while cleaning up

p2p/src/network.rs
Subhasish-Behera: Ah! Yep, the attestation is not required. ~The payload is also not required, as I am not setting up the block seen as true when the block arrives, and that would require calling the controller with the chain_link check, so I think this needs a bit of change~

fork_choice_control/src/events.rs
hangleang: @Subhasish-Behera I mean the `execution_payload_available` event emit when both payload and blob/data column sidecars are available, but at the point of payload accepted, there hasn't checked on data availability yet

fork_choice_store/src/error.rs
hangleang: you can't just change to arbitrary number without knowing what's it means

fork_choice_store/src/store.rs
hangleang: looked at the spec again, I don't think these checks are required since it already checked against during bid validation, so the only thing need to check is whether the envelope come from the same builder based on the accepted bid

Subhasish-Behera: ok, let me revisit the checks from spec.

fork_choice_store/src/error.rs
Subhasish-Behera: hmm. not wrapping around Arc/box is causing the likely increase.  4 uize = 24 + 8  bytes.    BlobSidecarOnIncorrectSubnet {
        blob_sidecar: Arc<BlobSidecar<P>>,
        expected: SubnetId,
        actual: SubnetId,
    }, arc = 8 + 2* 8 , then add enum tag value + roundoff = 32

fork_choice_control/src/events.rs
Subhasish-Behera: yeah. i changed that to use standard available event.

p2p/src/block_sync_service.rs
```diff
@@ -664,38 +663,20 @@ impl<P: Preset> BlockSyncService<P> {
...
-                                    payload_slot,
-                                );
-                                let slot_start_time = SystemTime::UNIX_EPOCH + Duration::from_secs(slot_start);
-                                let three_fourths_slot = slot_start_time + Duration::from_millis(
-                                    3 * self.config.seconds_per_slot.get() * 1000 / 4
+                                // Check if beacon block has been seen
+                                let beacon_block_seen = self.received_block_roots.contains_key(&beacon_block_root);
+
+                                debug!(
+                                    "received execution payload as gossip (slot: {payload_slot}, \
```
hangleang: actually this should called [`block_slot`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/builder.md?plain=1#L230)

fork_choice_control/src/mutator.rs
```diff
@@ -3390,7 +3390,7 @@ where
             mutator_tx: self.owned_mutator_tx(),
             wait_group,
             execution_payload_envelope,
-            beacon_block_seen,
+            beacon_block_seen: true,
```
hangleang: why are you so sure that the block has seen at the point of retry the payload?

fork_choice_store/src/store.rs
```diff
@@ -2444,6 +2444,7 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         envelope: Arc<SignedExecutionPayloadEnvelope<P>>,
         beacon_block_seen: bool,
         origin: &ExecutionPayloadEnvelopeOrigin,
+        state_fn: impl FnOnce() -> Option<Arc<BeaconState<P>>>,
```
hangleang: I should be more clear on the comment, you can access the state from ChainLink, which you already have [here](https://github.com/grandinetech/grandine/pull/483/commits/2734a3616c30ddb1d0ea80d1ce856db5692eb526#diff-de6907564e20082198dcffdff427589f3f0034ca4b09b6f269378a9e8d33fe5bL2468), so just call `.state`

fork_choice_control/src/tasks.rs
```diff
@@ -526,10 +526,13 @@ impl<P: Preset, W> Run for ExecutionPayloadEnvelopeTask<P, W> {
             .as_ref()
             .map(|metrics| metrics.fc_execution_payload_envelope_task_times.start_timer());
 
+        let beacon_block_root = execution_payload_envelope.message.beacon_block_root;
+
         let result = store_snapshot.validate_execution_payload_envelope(
             execution_payload_envelope,
             beacon_block_seen,
             &origin,
+            || store_snapshot.chain_link(beacon_block_root).map(|chain_link| chain_link.state(&store_snapshot)),
```
hangleang: ahah! that's the same loading the state

fork_choice_store/src/store.rs
```diff
@@ -2506,21 +2506,19 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+            envelope.signature,
+            self.pubkey_cache.get_or_insert(validator.pubkey)?,
+            SignatureKind::ExecutionPayloadEnvelope,
         )?;
 
         // [REJECT] Get the payload bid from state
         // Spec: "this can be obtained from the state.latest_execution_payload_bid"
-        let bid = &gloas_state.latest_execution_payload_bid;
+        let bid = match state.as_ref() {
+            BeaconState::Gloas(gloas_state) => gloas_state.latest_execution_payload_bid,
```
hangleang: you can call `state.post_gloas().latest_execution_payload_bid()` or `block.body().with_payload_bid().map(BlockBodyWithPayloadBid::signed_execution_payload_bid)` to get the bid, we don't want to add another match arm for every future phases if not necessary. you should also add the error to the type

Subhasish-Behera: makes sense

fork_choice_control/src/mutator.rs
Subhasish-Behera: changed the approach in [46ed9ad](https://github.com/grandinetech/grandine/pull/483/commits/46ed9ad47179bd655e485197e8630a2a5cc2ca38) . calling controller for pending envelopes on incoming block in block_sync_sevice.rs .

fork_choice_store/src/store.rs
```diff
@@ -2478,14 +2477,8 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
-        // Delay envelope validations until the state is available.
-        let Some(state) = state_fn() else {
-            return Ok(ExecutionPayloadEnvelopeAction::DelayUntilState(
-                envelope,
-                beacon_block_root,
-                slot,
-            ));
-        };
+        // Get state from chain_link (always available once chain_link exists)
+        let state = chain_link.state(self);
```
hangleang: :)

hangleang: there is a field [`ChainLink.state`](https://github.com/grandinetech/grandine/blob/0a9353e4b367961734d9e16aac3c7b005cafa432/fork_choice_store/src/misc.rs#L44), not only the [state](https://github.com/grandinetech/grandine/blob/0a9353e4b367961734d9e16aac3c7b005cafa432/fork_choice_store/src/misc.rs#L84-L86) function

p2p/src/block_sync_service.rs
```diff
@@ -100,7 +101,10 @@ pub struct BlockSyncService<P: Preset> {
     received_blob_sidecars: Arc<DashMap<BlobIdentifier, Slot>>,
     received_block_roots: HashMap<H256, Slot>,
     received_data_column_sidecars: Arc<DashMap<DataColumnIdentifier, Slot>>,
+    /// Deduplication: tracks seen envelopes by (block_root, builder_index) to prevent duplicate processing
     received_envelopes: HashMap<(H256, ValidatorIndex), Slot>,
+    /// Queue: envelopes waiting for block; avoids spawning tasks that would have not proceeded DelayUntilBeaconBlock
```
hangleang: why introduce a new flow for delayed envelopes, while we already have the `PendingExecutionPayloadEnvelope`?

Subhasish-Behera: i was trying not to spawn a task when a block is not present(handle the case of envelope present block abset at p2p layer). but taking another look , it seems like this skips the block `P2pMessage::BlockNeeded`

hangleang: the delay path won't spawn any task neither

Subhasish-Behera: I meant the delay flow would be
1. first task: `on_gossip_execution_payload` → `ExecutionPayloadEnvelopeTask`
2. run validation return `DelayUntilBeaconBlock`
3. mutator calls `delay_execution_payload_envelope_until_block` then `send_to_p2p(P2pMessage::BlockNeeded(beacon_block_root, peer_id))`
4. second task: when block arrives `retry_execution_payload_envelope` spawns second task
-----
in the handling of p2p approach(so only true task spawning happens once when block arrives), no immediate task spawning. but the problem is it cant be proactive in asking for blocks through BlockNeeded, and actually has to wait.

Subhasish-Behera: just removed pending envelopes

hangleang: @Subhasish-Behera thanks, should I review now, or wait until you rebase?

Subhasish-Behera: I was planning to squash and then rebase after all the commits once all reviews are resolved.

hangleang (approved):

PR #486, opened 2025-11-18
Handle execution payload bid

hangleang:
after #484


fork_choice_control/src/controller.rs
```diff
@@ -799,6 +820,33 @@ where
...
+        origin: ExecutionPayloadBidOrigin,
+    ) {
+        self.spawn_execution_payload_bid_task_with_wait_group(
+            self.owned_wait_group(),
+            payload_bid,
+            origin,
+        )
+    }
+
+    fn spawn_execution_payload_bid_task_with_wait_group(
```
povi: No need to have a separate `spawn_execution_payload_bid_task_with_wait_group` method if you don't plan to use it again somewhere else

fork_choice_control/src/thread_pool.rs
```diff
@@ -242,6 +244,12 @@ impl<P: Preset, E, W> Spawn<P, E, W> for PayloadAttestationTask<P, W> {
     }
 }
 
+impl<P: Preset, E, W> Spawn<P, E, W> for ExecutionPayloadBidTask<P, W> {
+    fn spawn(self, critical: &mut Critical<P, E, W>) {
+        critical.low_priority_tasks.push_back(self.into())
```
povi, line 250: I'm not sure if it should be low or mid priority task

fork_choice_store/src/store.rs
```diff
@@ -1278,6 +1279,99 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+    pub fn validate_execution_payload_bid(
+        &self,
+        payload_bid: Arc<SignedExecutionPayloadBid>,
+        origin: &ExecutionPayloadBidOrigin,
+    ) -> Result<ExecutionPayloadBidAction> {
+        let bid = payload_bid.message;
+        let builder_index = bid.builder_index;
+
+        // > off-protocol payment is disallowed in gossip, the `bid.execution_payment` MUST be zero
+        if origin.is_from_gossip() {
```
povi: What about bids from `POST /eth/v1/beacon/execution_payload_bid`? Are they allowed to have non-zero `execution_payment`?

povi (approved):

fork_choice_control/src/thread_pool.rs
hangleang: I will mark TODO there to figure out later

fork_choice_store/src/store.rs
hangleang: bid submitted from the API need to [pass the gossip validations](https://github.com/ethereum/beacon-APIs/pull/552/changes#diff-657fd447a2c8cb47c083391c360b840a3816ac08341669009e74beff2bf5f9a3), so it should also disallow

PR #487, opened 2025-11-18
Export telemetry metrics to the specified gRPC collector

Tumas:
- Trace block processing & import
- Add separate tracing filter
- Reload tracing filter through beacon API
- Compile core crates by default with tracing enabled. Use `make TRACING_DISABLED=1` to compile core modules (transition_functions, helper_functions) without tracing.

Collects tracing metrics only for root span by default. More detailed collection can be speicified via `--telemetry-level` CLI argument.

More fine-grained telemetry levels can be specified via `GRANDINE_TRACE_LOG` env variable or via beacon API.


povi (approved):
LGTM

PR #488, opened 2025-11-19
PTC Cache Implementation

Subhasish-Behera:
Implements caching for Payload Timeliness Committee (PTC) to avoid recomputing balance-weighted selection multiple times per epoch.

  **Problem:** PTC uses expensive balance-weighted selection. Without caching, the same committee is recomputed for every
  validator duty check and every attestation verification.

  **Solution:** Cache the entire epoch's PTC with constant time validator lookup.

  **New Functions:**
  - `get_validator_ptc_slots(validator_index)` - Returns all (slot, position) pairs where validator appears. 
  will be used in validator duty calls to know which slots to attest.
  - `get_ptc(state, slot)` - Returns full PTC committee for a slot. Used for gossip verification, block processing.


types/src/gloas/ptc_cache.rs
hangleang, line 1: can you simply explain what it does? what technique you use here to manage the cache? I can't fully wrap my head around

Subhasish-Behera: hey @hangleang so the goal is to reduce the calls and do the actual calculation once per epoch. 

So it operates like a regular cache. But the added element is that the `balance_weighted_selection`. So that's on top of seed and validator indices. 


  Call Flow

 ` get_ptc(state, slot)
    ↓

  state.ptc_cache().get_or_init()  ← OnceCell: thread-safe lazy init
    ↓
  build_ptc_cache(state, epoch)
    ↓
  for each slot in epoch:
      compute_ptc_for_slot_internal()  ← Calls balance_weighted_selection
    ↓
  Store in ptc_shuffling: [slot0: 512 validators, slot1: 512, ..., slot31: 512]
    ↓
  Build reverse indexes for validator→slots lookup `

so in first commit i made 2 mistakes 1. keeping a cache for 3 epochs but thats wrong as balance can change per epoch. 2. didnt handle multiple occurance of single validator for reverse lookup `ptc_positions`(used in `get_validator_ptc_slots` which answers which slots is this validator assigned).  

in 2nd commits. instead of keeping the track of validators position in a vec i kept in a vec<vec<>> to give multiple positons per validator. 

in 3rd commit , instead of creating vec vec for every validator, i changed to maintaining 2 lookups one`ptc_positions` for single occurance and other  `multiple_positions` for validators with multiple occurances. thus reducing the chances of 2 vector overhead for every single validator index. manged by a hash map with vec of position of slots  as value(validator index as key)

in 4th commit, i fixed my mistake of keeping/precalculating ptc for multiple epochs assuming that balance would remain same across multiple epochs.

the whole constraint of `build_ptc_cache` will be need once per epoch is managed by the cache clearing logic by calling `clear_ptc_cache` in `process_epoch`.
  1. Forward index (get_ptc): slot → validators
    - Uses ptc_shuffling (flat Vec)
    - "answers who are the ptc members for slot x"
  2. Reverse index (get_validator_ptc_slots): validator → slots
    - Uses ptc_positions + multiple_positions
    - "validator v is assigned to which slots"

hangleang: hey @Subhasish-Behera, I like the idea to cache PTC members for an epoch which can be used in [`validate_payload_attestation`](https://github.com/grandinetech/grandine/blob/epbs/fork_choice_store/src/store.rs#L2494). currently, we do have [`own_ptc_members.rs`](https://github.com/grandinetech/grandine/blob/epbs/validator/src/own_ptc_members.rs) which keep track of assigned slots for validators attached to the beacon node (_we don't need to keep track other validators assigned slots, which is non of the beacon node business_), so I think we don't need the reverse lookup per validator in this cache, this way we can make the cache less complex. I also don't think keep ptc cache within state is the right place, the best place might be in `fork_choice_store`

types/src/gloas/ptc_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+///
+/// - The cache is cleared at epoch boundaries because effective balances change
+///
+/// Stores the forward index (slot -> validators) as a flat array.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct PTCCache {
+    /// The epoch this cache was initialized for
+    pub initialized_epoch: Option<Epoch>,
+    /// Flat shuffling for all slots in epoch: [slot0_ptc..., slot1_ptc..., ...]
+    pub ptc_shuffling: Vec<ValidatorIndex>,
```
hangleang: why store in flat list rather than map by slot? also PTC indices should not be shuffled here. see https://jtraglia.github.io/eth-spec-viewer/#nightly/functions-get_ptc

types/src/gloas/ptc_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// PTC size per slot
+    pub ptc_size: usize,
+    /// Slots per epoch
+    pub slots_per_epoch: u64,
```
hangleang: both values can be derived from `Preset`, as commented above if store in map, then no need to chunk array, also those values no longer needed

types/src/gloas/beacon_state.rs
```diff
@@ -123,4 +123,10 @@ pub struct BeaconState<P: Preset> {
     #[serde(skip)]
     #[ssz(skip)]
     pub cache: Cache,
+
+    // PTC cache
+    #[derivative(PartialEq = "ignore")]
+    #[serde(skip)]
+    #[ssz(skip)]
+    pub ptc_cache: PtcCache,
```
hangleang: I think we can extend existing [cache](https://github.com/grandinetech/grandine/blob/develop/types/src/cache.rs) mechanism with tracking PTC indices for relative slots, and extend its `advance_slot` to shift the cache during slot processing.

Keep in mind that the `get_ptc` function doesn't require `PostGloasBeaconState`

types/src/gloas/ptc_cache.rs
Subhasish-Behera: coming to think of it, i didn't reconsider the structure after removing ptc_positions. vec was simpler earlier as the ptc_positions(reverse index) was also a flat vec. so yep, map makes more sense now. 
> should not be shuffled here

right, kept the name as other caches were using shuffled.  will change that.

helper_functions/src/accessors.rs
```diff
@@ -1117,35 +1096,80 @@ fn compute_ptc_for_slot_internal<P: Preset>(
...
+        let relative_slot = if slot + 1 == state_slot {
+            RelativeSlot::Previous
+        } else if slot == state_slot {
+            RelativeSlot::Current
+        } else if slot == state_slot + 1 {
+            RelativeSlot::Next
+        } else {
+            // Outside cache range - compute on demand (shouldn't happen per spec)
+            return ptc_for_slot(state, slot);
+        };
```
hangleang: you can follow https://github.com/grandinetech/grandine/blob/8ac065da176067bc4eb8b79ebfc48a6433f52499/helper_functions/src/accessors.rs#L95

types/src/cache.rs
```diff
@@ -61,8 +63,12 @@ impl Cache {
...
         balance[RelativeEpoch::Current] = core::mem::take(&mut balance[RelativeEpoch::Next]);
 
-        // Clear PTC cache (no-op for non-Gloas since never initialized)
-        ptc.take();
+        // Clear PTC cache - balance_weighted_selection depends on effective balances
+        // which change at epoch boundaries
+        let ptc = &mut self.ptc_cache;
+        ptc[RelativeSlot::Previous] = OnceCell::new();
+        ptc[RelativeSlot::Current] = OnceCell::new();
+        ptc[RelativeSlot::Next] = OnceCell::new();
```
hangleang: PTC cache is shift by slot basis as in `advance_slot`, it also pre computed in preprocess state task

fork_choice_control/src/tasks.rs
```diff
@@ -886,6 +886,10 @@ fn initialize_preprocessed_state_cache<P: Preset>(
     accessors::get_or_init_total_active_balance(state, false);
     accessors::get_or_init_validator_indices(state, false);
 
+    // Pre-compute PTC cache (no-op for non-Gloas)
+    accessors::get_or_try_init_ptc(state, RelativeSlot::Current, false)?;
+    accessors::get_or_try_init_ptc(state, RelativeSlot::Next, false)?;
```
hangleang, line 891: here

helper_functions/src/accessors.rs
```diff
@@ -1117,35 +1096,80 @@ fn compute_ptc_for_slot_internal<P: Preset>(
...
+            .ptc_cache[relative_slot]
+            .get_or_try_init(|| {
+                if report_cache_miss {
+                    #[cfg(feature = "metrics")]
+                    if let Some(metrics) = METRICS.get() {
+                        metrics.ptc_cache_init_count.inc();
+                    }
+                }
+                compute_ptc_for_slot_internal(state, slot)
+            })?;
```
hangleang, line 1143: why don't return its result?

helper_functions/src/accessors.rs
```diff
@@ -1117,35 +1096,80 @@ fn compute_ptc_for_slot_internal<P: Preset>(
...
+
+        // Initialize cache if needed
+        get_or_try_init_ptc(state, relative_slot, true)?;
+
+        // Get the cached value
+        state
+            .cache()
+            .ptc_cache[relative_slot]
+            .get()
+            .expect("just initialized by get_or_try_init_ptc")
```
hangleang: cached could be get from `get_or_try_init_ptc`

helper_functions/src/accessors.rs
```diff
@@ -1117,35 +1096,80 @@ fn compute_ptc_for_slot_internal<P: Preset>(
...
+            .iter()
+            .copied()
+            .pipe(ContiguousVector::try_from_iter)
+            .map_err(Into::into)
+    } else {
+        ptc_for_slot(state, slot)
```
hangleang: does PTC needed for pre-Gloas? I think we don't need to check phase here

types/src/cache.rs
Subhasish-Behera: but the `effective_balance` changes at epoch boundry, i.e the `balance_weighted_selection`. if we dont clear it here, then at slot 31: next = `ptc(32)` -> `advance_slot()`: Curr = `ptc(32, old balances)` -> `process_epoch`: balance changed -> `advance_epoch`: Curr has old balance based ptc. -> goes to `initialize_preprocessed_state_cache` and calls `get_or_try_init_ptc` with old value only as `OneCell` wont recompute as value already present.

hangleang: I might be wrong, but the spec stated:

> A validator may be a member of the new Payload Timeliness Committee (PTC) for a given slot. To check for PTC assignments, use get_ptc_assignment(state, epoch, validator_index) where epoch <= next_epoch, **as PTC committee selection is only stable within the context of the current and next epoch**.

Subhasish-Behera: What I interpret from the statement is: here it is talking about `get_seed_by_epoch`. Yes it is stable, because `compute_balance_weighted_selection` uses `get_seed_by_epoch` which uses `get_randao_mix`. so   
 ```rust
  fn get_seed_by_epoch<P: Preset>(..., epoch: Epoch, ...) -> H256 {
      let mix = get_randao_mix(state, epoch - MIN_SEED_LOOKAHEAD - 1);  // epoch - 2
      hash_32_64_256(domain_type, epoch, mix)
  }
  ```
so stability in terms of what you will get if you query. 
current = epoch N = query indirectly `get_seed_by_epoch` is stable 
same for next = epoch N+ 1
beyond that i.e epoch N+2, the randao reveal is accumulating via process_randao(), hence not stable. i.e `get_seed_by_epoch`will likely vary from `N+2 slot x` vs `N+2 slot y` .  so stablity at current and next epoch not same stability across current and next epoch

Subhasish-Behera: we can ask povilas to clear things up if required

hangleang (approved):
Look good to me, except the PTC cache prune after epoch transition. that would need @povi to take a look

povi (approved):
LGTM

PR #494, opened 2025-11-24
feat: enable internal block building route for gloas

Subhasish-Behera:
Add complete self-build flow for Gloas phase where validator creates execution payload bid, embeds it in beacon block, and publishes signed execution payload envelope separately.

Validator responsibilities:
- Request block producer to fetch payload and create unsigned bid
- Sign ExecutionPayloadBid and pass to block producer
- Sign RANDAO reveal for block building
- Sign final BeaconBlock
- Retrieve cached envelope_data from block producer
- Build and sign ExecutionPayloadEnvelope
- Publish signed block and envelope via P2P

Block producer responsibilities:
- Fetch execution payload from EL (single fetch)
- Create unsigned ExecutionPayloadBid from payload
- Cache envelope_data (payload, execution_requests, commitments)
- Embed signed bid into block body
- Build block with state root
- Provide cached envelope_data to validator on request

roughly follows this flow
``` [V] ProposeBlock (Self-Build)
   ├─ [V→BN] build_execution_payload_bid_and_cache_envelope_data
   │   ├─ [BN→EL] notify_forkchoice_updated → payload_id
   │   ├─ [BN→EL] get_execution_payload(payload_id) → ExecutionPayload
   │   ├─ [BN] prepare_execution_payload_bid
   │   └─ [BN] cache envelope_data
   ├─ [V] sign(ExecutionPayloadBid) → SignedExecutionPayloadBid
   ├─ [V→BN] build_blinded_beacon_block(randao_reveal, signed_bid)
   │   └─ [BN] prepare_gloas_block_with_bid (embed signed_bid)
   ├─ [V] sign(BeaconBlock) → SignedBeaconBlock
   ├─ [V→BN] PublishBeaconBlock
   │   └─ [P2P] gossip SignedBeaconBlock
   └─ [V] SubmitEnvelope
       ├─ [V→BN] get_cached_envelope_data
       ├─ [V] sign(ExecutionPayloadEnvelope) → SignedEnvelope
       └─ [V→P2P] PublishExecutionPayloadEnvelope
           └─ [P2P] publish_execution_payload_envelope ```

### DIFF
--- a/block_producer/src/block_producer.rs
+++ b/block_producer/src/block_producer.rs
@@ -105,7 +105,7 @@ const PAYLOAD_ID_CACHE_SIZE: usize = 10;
 
 pub type ExecutionPayloadHeaderJoinHandle<P> = JoinHandle<Result<Option<SignedBuilderBid<P>>>>;
 pub type LocalExecutionPayloadJoinHandle<P> =
-    JoinHandle<Option<(WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>, H256)>>;
+    JoinHandle<Option<WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>>>;
 
 type PayloadCache<P> =
     Mutex<SizedCache<H256, WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>>>;
@@ -1152,27 +1152,20 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
         let can_self_build = !self.beacon_state.is_post_gloas()
             || predicates::has_builder_withdrawal_credential(self.proposer()?);
 
-        let (payload_with_data, payload_root) = if can_self_build {
+        let payload_with_data = if can_self_build {
             if let Some(handle) = local_execution_payload_handle {
-                match handle.await? {
-                    Some((payload, root)) => (
-                        Some(payload.map(|value| value.map(Some))),
-                        Some(root),
-                    ),
-                    None => (None, None),
-                }
+                handle
+                    .await?
+                    .map(|value| value.map(|value| value.map(Some)))
             } else {
-                (None, None)
+                None
             }
         } else {
             // To prevent return error in payload_with_data check
-            (
-                Some(WithClientVersions {
-                    client_versions: Some(vec![ClientVersionV1::own()].into()),
-                    result: WithBlobsAndMev::with_default(None),
-                }),
-                None,
-            )
+            Some(WithClientVersions {
+                client_versions: Some(vec![ClientVersionV1::own()].into()),
+                result: WithBlobsAndMev::with_default(None),
+            })
         };
 
         let WithClientVersions {
@@ -1200,9 +1193,11 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
         };
 
         // Cache payload root for Gloas envelope retrieval later (only when self-building)
-        // payload_root already computed in local_execution_payload_result, no need to recompute
+        // External builders publish their own envelope
         if can_self_build && self.beacon_state.phase() >= Phase::Gloas {
-            *self.cached_payload_root.lock().await = payload_root;
+            if let Some(ref payload) = execution_payload {
+                *self.cached_payload_root.lock().await = Some(payload.hash_tree_root());
+            }
         }
 
         let mut without_state_root = if let Some(state) = self.beacon_state.post_gloas() {
@@ -1978,7 +1973,7 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
 
     async fn local_execution_payload_result(
         &self,
-    ) -> Result<Option<(WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>, H256)>> {
+    ) -> Result<Option<WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>>> {
         let snapshot = self.producer_context.controller.snapshot();
 
         let mut payload_id = self
@@ -2073,16 +2068,15 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
             .await
             .cache_set(payload_root, payload.clone());
 
-        Ok(Some((payload, payload_root)))
+        Ok(Some(payload))
     }
 
     // If the local execution engine fails, a block can still be constructed with a payload received
     // from an external block builder or even the default payload, though blocks with default
     // payloads are only valid before the Merge.
-    // Returns (payload, payload_root) to avoid recomputing hash_tree_root
     async fn local_execution_payload_option(
         &self,
-    ) -> Option<(WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>, H256)> {
+    ) -> Option<WithClientVersions<WithBlobsAndMev<ExecutionPayload<P>, P>>> {
         if self.producer_context.fake_execution_payloads {
             let slot = self.beacon_state.slot();
 
@@ -2101,11 +2095,9 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
 
                 match execution_payload {
                     Ok(payload) => {
-                        let payload_root = payload.hash_tree_root();
-                        return Some((
-                            WithClientVersions::none(WithBlobsAndMev::with_default(payload)),
-                            payload_root,
-                        ))
+                        return Some(WithClientVersions::none(WithBlobsAndMev::with_default(
+                            payload,
+                        )))
                     }
                     Err(error) => panic!("failed to produce fake payload: {error:?}"),
                 };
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: block_producer/src/block_producer.rs [hunk 1]
```diff
@@ -638,20 +641,30 @@ pub struct BlockBuildOptions {
     pub builder_boost_factor: Uint256,
 }
 
+/// Data needed to construct ExecutionPayloadEnvelope later (after signing block)
+#[derive(Clone)]
+pub struct EnvelopeData<P: Preset> {
```
hangleang: those data are available in [`payload_cache`](https://github.com/grandinetech/grandine/blob/0a9353e4b367961734d9e16aac3c7b005cafa432/block_producer/src/block_producer.rs#L1822-L1826) after call `engine_getPayload`

block_producer/src/block_producer.rs
```diff
@@ -1173,10 +1189,19 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
...
 
-        let mut without_state_root_with_payload = block_without_state_root
-            .with_execution_payload(execution_payload)?
-            .with_blob_kzg_commitments(commitments)
-            .with_execution_requests(execution_requests);
+        let mut without_state_root_with_payload = if self.beacon_state.phase() >= Phase::Gloas
+            && signed_bid.is_some()
+        {
+            // Gloas self-build: Use cached envelope data (don't use execution_payload to avoid double fetch)
+            let signed_bid = signed_bid.expect("signed_bid checked above");
```
hangleang: if that's the case, why don't use `if let`?

block_producer/src/block_producer.rs
```diff
@@ -1187,6 +1192,14 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
             }
         };
 
+        // Cache payload root for Gloas envelope retrieval later (only when self-building)
+        // External builders publish their own envelope
+        if can_self_build && self.beacon_state.phase() >= Phase::Gloas {
+            if let Some(ref payload) = execution_payload {
+                *self.cached_payload_root.lock().await = Some(payload.hash_tree_root());
+            }
```
hangleang, line 1204: this block (cache payload root) can be put before https://github.com/grandinetech/grandine/blob/05d70b9815692c89a91ca07dfc8a5ad99d8169c3/block_producer/src/block_producer.rs#L1192
and if possible try to expose the computed hash root to avoid re-compute root of the payload https://github.com/grandinetech/grandine/blob/05d70b9815692c89a91ca07dfc8a5ad99d8169c3/block_producer/src/block_producer.rs#L1963

block_producer/src/block_producer.rs
```diff
@@ -648,6 +651,8 @@ pub struct BlockBuildContext<P: Preset, W: Wait> {
     head_block_root: H256,
     proposer_index: ValidatorIndex,
     options: BlockBuildOptions,
+    /// Payload root for retrieving cached envelope data from payload_cache
+    cached_payload_root: Arc<Mutex<Option<H256>>>,
```
hangleang:
```suggestion
    cached_payload_root: Option<H256>,
```
I see no race condition here for a single root

validator/src/validator.rs
```diff
@@ -1139,7 +1140,83 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
                 self.controller
                     .on_own_block(wait_group.clone(), block.clone_arc());
 
-                ValidatorToP2p::PublishBeaconBlock(block).send(&self.p2p_tx);
+                ValidatorToP2p::PublishBeaconBlock(block.clone_arc()).send(&self.p2p_tx);
+
+                // Handle Gloas execution payload envelope (only for self-build)
+                if let Some((deneb_payload, execution_requests_data, blob_kzg_commitments)) =
```
hangleang: put this block into a function, and call it

validator/src/validator.rs
```diff
@@ -2037,7 +2114,7 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
 
             let doppelganger_protection = self
                 .doppelganger_protection
-                .as_deref()
+                .as_deref()
```
hangleang, line 2144: fix lint

transition_functions/src/lib.rs
```diff
@@ -246,7 +247,7 @@ pub mod gloas {
     mod block_processing;
     mod epoch_intermediates;
     mod epoch_processing;
-    mod execution_payload_processing;
+    pub mod execution_payload_processing;
```
hangleang: do you need to export this module?

validator/src/validator.rs
```diff
@@ -1139,7 +1140,83 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
...
+                        beacon_block_root,
+                        slot: slot_head.slot(),
+                        blob_kzg_commitments,
+                        state_root,
+                    };
+
+                    // Sign the envelope
+                    let envelope_signing_root = envelope.signing_root(&self.chain_config, &slot_head.beacon_state);
+                    let envelope_sig = match signer_snapshot
+                        .sign_without_slashing_protection(
```
hangleang: follow `sign_beacon_block` function

hangleang: let's me check with Povi

block_producer/src/block_producer.rs
```diff
@@ -1871,6 +1884,95 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
...
+        ))
+    }
+
+    /// Compute the post-execution state root for the envelope
+    pub async fn compute_post_execution_state_root(
+        &self,
+        beacon_block_root: H256,
+        builder_index: ValidatorIndex,
+    ) -> Result<H256> {
+        let (deneb_payload, execution_requests_data, blob_kzg_commitments) = self
```
hangleang: can these values be passed from params in `validator.rs`?

block_producer/src/block_producer.rs
```diff
@@ -1871,6 +1884,95 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
...
+            blob_kzg_commitments,
+            state_root: H256::zero(), // Placeholder
+        };
+
+        let signed_envelope = SignedExecutionPayloadEnvelope {
+            message: envelope,
+            signature: SignatureBytes::default(),
+        };
+
+        let BeaconState::Gloas(ref gloas_state) = *self.beacon_state else {
```
hangleang: should call `post_gloas` instead

Subhasish-Behera: we are cloning the state to pass it to `process_execution_payload` which mutates the state. `post_gloas` will return a trait of `PostGloasBeaconState` which we cant clone.
(and we have to clone to mutate as beacon_state in BlockBuildContext is `Arc`. and even if we do `Arc<Mutex>` field for `beacon_state` by which we can pass state without clone, but rest of the places using beacon_state need pre `proces_execution_payload` state ).

block_producer/src/block_producer.rs
Subhasish-Behera: I am not sure if plain `Option` will work here because, because the value of cached_payload_root is being mutated as well as read with `&self` i.e start with `None` at `new_build_context` (`cached_payload_root: Arc::new(Mutex::new(None))`) -> then actual value assigned at `produce_beacon_block` (1) and read at `get_gloas_envelope_data`(2). both at 1,2 `cached_payload_root` is accesed via `*self` . without `Arc<Mutex>` i think it would need `&mut self`  .  (tldr: builder context created before payload creation, hence required mutation and mutation is interior as done through `&self`)

- Though I tried using `Cell<Option<H256>>`, but that wont work here as there are async block moving builder_context, but Cell is not `Sync`. 
- Option worked with `options: BlockBuildOptions 's graffiti field` as its not mutated

validator/src/validator.rs
hangleang: he said it is fine

block_producer/src/block_producer.rs
hangleang: can you revert this commit https://github.com/grandinetech/grandine/pull/494/changes/5f871efdb0ebd9f9a8d8849a890810bed0394f3b with `git revert 5f871efd`? I don't expect to change those existing functions signature

block_producer/src/block_producer.rs
hangleang: you are right, just keep as it

block_producer/src/block_producer.rs
hangleang: the problem with this control statement is when we add new phase, we often unaware of this check to add another `else if`, which lead to bug return error. so for now, pls change this to match arm, we will refactor it later to match the function signature

block_producer/src/block_producer.rs
Subhasish-Behera: ok. I thought by "expose" you meant passing through params

hangleang (dismissed):

PR #502, opened 2025-12-01
Add Forkchoice changes for ePBS

Subhasish-Behera:
Major changes:
The whole logic is based on the idea that it doesn't matter how you reach the segment-level identifiers. i.e though ePBS introduces a separate envelope object, but the same segment structure can be followed. So the difference between a envelope(full block)/ empty block is which state is read from that block. One of the ways to do that is by following the existing pattern of keeping a chunk of the Store fields inside Chainlink
1. ChainLink dual statefield
```
  block_state: Option<Arc<BeaconState<P>>>,           // Pre-execution (after beacon block STF)
  execution_payload_state: Option<Arc<BeaconState<P>>>, // Post-execution (after payload STF)
```
The changed structure of empty/full block pretty much behave as siblings if it is looked from view of part of forkchoice calculation which depends on reading parent information. 
- with epbs, to get correct forchoice results, its important to know both which variant the block itself is + which block will it be effecting(for eg: not just the insertion of envelope/empty block have a type but also what it is inserted into has a type  i.e the parent block can be full as well as empty)
so the the most common structure that is going to happen is:
```
 Segment 0:                ──► [Parent X]
                                │
                                ├──► Segment 1: [Empty A] ──► [Empty B]
                                │                  ▲
                                │                  │
                                │    unfinalized_locations_empty["A"] = (seg:1, pos:0)
                                │
                                └──► Segment 2: [Full A]
                                                 ▲
                                                 │
                                     unfinalized_locations_full["A"] = (seg:2, pos:0)
```
- both weight propagation purposes as well as the insertion purpose of full/empty block of arrival, the choice of introducing dual maps is much easier than differentiation at `unfinalized_block` level or using `chain_link` fields.
-  the key currently for both the map is `block_root`. A previous approach was made to keep unfinalized_locations_full maps key as `execution_hash` and unfinalized_locations_empty through `block_root`. But there was significant overhead  in terms of number of conversion calla/reverse lookup for hash to root and viceversa. plus as only one variant of full/empty per blockroot can reach till forkchoice, there is ideally no problem with having the same keys.  

- a way to move forward with the changes is to look at in which scenarios(insertion into segment, calculation supporting vote, tiebraker , weight propagtion/childern filter/setting `ForkChoicePayloadStatus`) empty vs full is preferred across spec and current code.


fork_choice_control/src/mutator.rs
```diff
@@ -624,6 +629,7 @@ where
     ) -> Result<()> {
         match *result {
             Ok(BlockAction::Accept(mut chain_link, attester_slashing_results)) => {
+                eprintln!("DEBUG: BlockAction::Accept for root={:?}", block_root);
```
povi: Remove `eprintln!` from PR, if you want to add permanent debug info to fork choice, use `trace_with_peers!`

fork_choice_store/src/store.rs
```diff
@@ -470,20 +486,43 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
 
+    /// ePBS: Get ChainLink for block_root, preferring full variant if both exist.
+    ///
+    /// Used in validation flow to select parent state:
+    /// - Full variant has execution_payload_state (post-execution) → preferred
+    /// - Empty variant has only block_state (pre-execution) → fallback
+    ///
+    /// This uses get_location() which prefers full variant when both exist.
+    #[must_use]
+    pub fn chain_link_prefer_full(&self, block_root: H256) -> Option<&ChainLink<P>> {
```
povi: Method body is the same as `Store::chain_link`. Remove the duplicate

fork_choice_store/src/store.rs
```diff
@@ -502,7 +541,272 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+    /// - Fork choice operations prefer complete variant (full has execution_state)
+    /// - Children of block consistently extend from same variant
+    ///
+    /// Lookup strategy (optimized with block_root keying):
+    /// 1. Check unfinalized_locations_full by block_root (prefer full if exists)
+    /// 2. Otherwise check unfinalized_locations_empty by block_root
+    /// 3. Both maps keyed by same beacon_block_root, so no conversion needed
+    #[must_use]
+    fn get_location(&self, block_root: H256) -> Option<Location> {
+        // Check full variant first (prefer it if exists)
```
povi: Can we do that? What if validators vote for node with no payload and the node with no payload has bigger weight?

fork_choice_store/src/store.rs
povi: Code still contains duplicate methods with different names

fork_choice_store/src/store.rs
```diff
@@ -735,6 +1179,81 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+
+    /// [`is_data_available`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.1/specs/fulu/fork-choice.md#modified-is_data_available)
+    #[must_use]
+    pub fn is_data_available(&self, block_root: H256) -> bool {
+        let Some(chain_link) = self.chain_link(block_root) else {
+            return false;
+        };
+
+        if chain_link.execution_payload_state.is_none() {
+            return true;
```
povi: This code block does nothing

fork_choice_store/src/store.rs
```diff
@@ -4007,28 +5236,82 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+                    let branch_tiebreaker = self.get_payload_status_tiebreaker(branch_root, branch_status);
+
+
+                    // Compare (root, tiebreaker) per spec - lower values lose
+                    if (sibling_root, sibling_tiebreaker) < (branch_root, branch_tiebreaker) {
+                        best_descendant_of_segment = Some(branch_point.best_descendant);
+                    } else {
+                    }
+                    // else: sibling wins or equal (prefer sibling on exact tie)
+                } else {
```
povi: What does this `else` do here?

fork_choice_control/src/queries.rs
```diff
@@ -977,6 +995,24 @@ where
             .chain_link(block_root)
             .map(|chain_link| chain_link.payload_status)
     }
+
+    /// Check if execution payload envelope has been seen for block_root.
+    #[must_use]
+    pub fn has_envelope(&self, block_root: H256) -> bool {
```
povi: Duplicate method definition in `Controller`

fork_choice_control/src/controller.rs
```diff
@@ -983,6 +983,21 @@ where
         self.store_snapshot().sampling_columns_count()
     }
 
+    /// Check if execution payload envelope has been seen for block_root.
+    pub fn has_envelope(&self, block_root: H256) -> bool {
```
povi, line 983: Duplicate method definition in `Controller`

fork_choice_control/src/controller.rs
```diff
@@ -983,6 +983,21 @@ where
...
+        self.store_snapshot().has_envelope(block_root)
+    }
+
+    /// Check if blob data is available for block_root.
+    pub fn is_data_available(&self, block_root: H256) -> bool {
```
povi, line 988: Duplicate method definition in `Controller`

fork_choice_control/src/controller.rs
```diff
@@ -983,6 +983,21 @@ where
...
+        self.store_snapshot().is_data_available(block_root)
+    }
+
+    /// Get the slot of a block by its root.
+    pub fn block_slot(&self, block_root: H256) -> Option<Slot> {
```
povi, line 993: Duplicate method definition in `Controller`

fork_choice_control/src/queries.rs
```diff
@@ -977,6 +995,24 @@ where
...
+        self.store_snapshot().has_envelope(block_root)
+    }
+
+    /// Check if blob data is available for block_root.
+    #[must_use]
+    pub fn is_data_available(&self, block_root: H256) -> bool {
```
povi: Duplicate method definition in `Controller`

fork_choice_control/src/queries.rs
```diff
@@ -977,6 +995,24 @@ where
...
+        self.store_snapshot().is_data_available(block_root)
+    }
+
+    /// Get the slot of a block by its root.
+    #[must_use]
+    pub fn block_slot(&self, block_root: H256) -> Option<Slot> {
```
povi: Duplicate method definition in `Controller`

fork_choice_control/src/spec_tests.rs
```diff
@@ -347,6 +384,18 @@ fn run_case<P: Preset>(config: &Arc<Config>, case: Case) {
                 if let Some(proposer_boost_root) = proposer_boost_root {
                     context.assert_proposer_boost_root(proposer_boost_root);
                 }
+
+                if let Some(count) = execution_payload_states_count {
+                    context.assert_execution_payload_states_count(count);
```
povi: Method does not exist anywhere

fork_choice_control/src/spec_tests.rs
```diff
@@ -347,6 +384,18 @@ fn run_case<P: Preset>(config: &Arc<Config>, case: Case) {
...
+                }
+
+                if let Some(count) = blocks_with_ptc_votes {
+                    context.assert_blocks_with_ptc_votes_count(count);
+                }
+
+                if let Some(status) = head_payload_status {
+                    context.assert_head_payload_status(status);
```
povi: Method does not exist anywhere

fork_choice_control/src/spec_tests.rs
```diff
@@ -347,6 +384,18 @@ fn run_case<P: Preset>(config: &Arc<Config>, case: Case) {
                 if let Some(proposer_boost_root) = proposer_boost_root {
                     context.assert_proposer_boost_root(proposer_boost_root);
                 }
+
+                if let Some(count) = execution_payload_states_count {
+                    context.assert_execution_payload_states_count(count);
+                }
+
+                if let Some(count) = blocks_with_ptc_votes {
+                    context.assert_blocks_with_ptc_votes_count(count);
```
povi: Method does not exist anywhere

PR #517, opened 2025-12-04
`POST /eth/v1/beacon/execution_payload/bid`

hangleang:
consider this PR after #486


http_api/src/misc.rs
```diff
@@ -333,6 +333,35 @@ impl<'de, P: Preset> DeserializeSeed<'de> for SingleApiAttestationListPhaseDeser
     }
 }
 
+pub struct SignedExecutionPaylodBidPhaseDeserializer<P> {
```
Tumas: typo: should be 'Payload'

PR #518, opened 2025-12-04
`GET /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}`

hangleang:


http_api/src/standard.rs
```diff
@@ -3278,6 +3279,34 @@ pub async fn validator_sync_committee_contribution<P: Preset, W: Wait>(
...
+/// `GET /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}`
+pub async fn validator_execution_payload_bid<P: Preset, W: Wait>(
+    State(controller): State<ApiController<P, W>>,
+    EthPath(slot): EthPath<Slot>,
+    EthPath(builder_index): EthPath<BuilderIndex>,
+    headers: HeaderMap,
+) -> Result<EthResponse<ExecutionPayloadBid<P>, (), JsonOrSsz>, Error> {
+    let block_root = controller.head().value.block_root;
+    let beacon_state = controller
+        .preprocessed_state_for_block_production(block_root, slot)
```
Tumas: API spec describes the slot parameter as: _Slot for which the execution payload bid is requested. **Must be current slot or next slot**_. 

This property of (current or next slot) should be validated. Then you can call `preprocessed_state_at_current_slot` or `preprocessed_state_at_next_slot` (the latter method should be added). 

This would eliminate the potentially wasteful and unnecessary state processing in case slot parameter is invalid.

hangleang: updated in https://github.com/grandinetech/grandine/pull/518/commits/b168b89191586217a208c6e98dd89909290e7032

Tumas (approved):

PR #523, opened 2025-12-09
Feature/disk usage optimization

galadd:
This PR implements delta encoding for beacon state storage, reducing 
disk usage by ~80% while significantly improving state I/O performance.

- 80% disk usage reduction (> 90% in reduction from full state to delta state)
- >70% faster state reads and writes (with cached base)

### DIFF
--- a/fork_choice_control/src/delta/mod.rs
+++ b/fork_choice_control/src/delta/mod.rs
@@ -1,0 +1,124 @@
+use serde::{Deserialize, Serialize};
+use ssz::Hc;
+use std::sync::Arc;
+use types::{combined::BeaconState, phase0::primitives::Slot, preset::Preset};
+
+mod altair;
+mod bellatrix;
+mod capella;
+mod deneb;
+mod electra;
+mod fulu;
+mod phase0;
+
+pub use altair::{
+    apply_delta as apply_altair_delta, delta as altair_delta,
+    BeaconStateDelta as AltairBeaconStateDelta,
+};
+pub use bellatrix::{
+    apply_delta as apply_bellatrix_delta, delta as bellatrix_delta,
+    BeaconStateDelta as BellatrixBeaconStateDelta,
+};
+pub use capella::{
+    apply_delta as apply_capella_delta, delta as capella_delta,
+    BeaconStateDelta as CapellaBeaconStateDelta,
+};
+pub use deneb::{
+    apply_delta as apply_deneb_delta, delta as deneb_delta,
+    BeaconStateDelta as DenebBeaconStateDelta,
+};
+pub use electra::{
+    apply_delta as apply_electra_delta, delta as electra_delta,
+    BeaconStateDelta as ElectraBeaconStateDelta,
+};
+pub use fulu::{
+    apply_delta as apply_fulu_delta, delta as fulu_delta, BeaconStateDelta as FuluBeaconStateDelta,
+};
+pub use phase0::{
+    apply_delta as apply_phase0_delta, delta as phase0_delta,
+    BeaconStateDelta as Phase0BeaconStateDelta,
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "", deny_unknown_fields)]
+pub enum BeaconStateDelta<P: Preset> {
+    Phase0(Phase0BeaconStateDelta<P>),
+    Altair(AltairBeaconStateDelta<P>),
+    Bellatrix(BellatrixBeaconStateDelta<P>),
+    Capella(CapellaBeaconStateDelta<P>),
+    Deneb(DenebBeaconStateDelta<P>),
+    Electra(ElectraBeaconStateDelta<P>),
+    Fulu(FuluBeaconStateDelta<P>),
+}
+
+impl<P: Preset> BeaconStateDelta<P> {
+    pub fn slot(&self) -> Slot {
+        match self {
+            BeaconStateDelta::Phase0(phase0_delta) => phase0_delta.slot,
+            BeaconStateDelta::Altair(altair_delta) => altair_delta.slot,
+            BeaconStateDelta::Bellatrix(bellatrix_delta) => bellatrix_delta.slot,
+            BeaconStateDelta::Capella(capella_delta) => capella_delta.slot,
+            BeaconStateDelta::Deneb(deneb_delta) => deneb_delta.slot,
+            BeaconStateDelta::Electra(electra_delta) => electra_delta.slot,
+            BeaconStateDelta::Fulu(fulu_delta) => fulu_delta.slot,
+        }
+    }
+}
+
+pub fn delta<P: Preset>(
+    base: Arc<BeaconState<P>>,
+    target: Arc<BeaconState<P>>,
+) -> BeaconStateDelta<P> {
+    match (base.as_ref(), &target.as_ref()) {
+        (BeaconState::Phase0(base), BeaconState::Phase0(target)) => {
+            BeaconStateDelta::Phase0(phase0_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Altair(base), BeaconState::Altair(target)) => {
+            BeaconStateDelta::Altair(altair_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Bellatrix(base), BeaconState::Bellatrix(target)) => {
+            BeaconStateDelta::Bellatrix(bellatrix_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Capella(base), BeaconState::Capella(target)) => {
+            BeaconStateDelta::Capella(capella_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Deneb(base), BeaconState::Deneb(target)) => {
+            BeaconStateDelta::Deneb(deneb_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Electra(base), BeaconState::Electra(target)) => {
+            BeaconStateDelta::Electra(electra_delta((**base).clone(), (**target).clone()))
+        }
+        (BeaconState::Fulu(base), BeaconState::Fulu(target)) => {
+            BeaconStateDelta::Fulu(fulu_delta((**base).clone(), (**target).clone()))
+        }
+        _ => panic!("Cannot create delta across different phases"),
+    }
+}
+
+pub fn apply_delta<P: Preset>(base: BeaconState<P>, delta: BeaconStateDelta<P>) -> BeaconState<P> {
+    match (&base, delta) {
+        (BeaconState::Phase0(base), BeaconStateDelta::Phase0(delta)) => {
+            BeaconState::Phase0(Hc::from(apply_phase0_delta((**base).clone(), delta)))
+        }
+        (BeaconState::Altair(base), BeaconStateDelta::Altair(delta)) => {
+            BeaconState::Altair(Hc::from(apply_altair_delta((**base).clone(), delta)))
+        }
+        (BeaconState::Bellatrix(base), BeaconStateDelta::Bellatrix(delta)) => {
+            BeaconState::Bellatrix(Hc::from(apply_bellatrix_delta((**base).clone(), delta)))
+        }
+        (BeaconState::Capella(base), BeaconStateDelta::Capella(delta)) => {
+            BeaconState::Capella(Hc::from(apply_capella_delta((**base).clone(), delta)))
+        }
+        (BeaconState::Deneb(base), BeaconStateDelta::Deneb(delta)) => {
+            BeaconState::Deneb(Hc::from(apply_deneb_delta((**base).clone(), delta)))
+        }
+
+        (BeaconState::Electra(base), BeaconStateDelta::Electra(delta)) => {
+            BeaconState::Electra(Hc::from(apply_electra_delta((**base).clone(), delta)))
+        }
+        (BeaconState::Fulu(base), BeaconStateDelta::Fulu(delta)) => {
+            BeaconState::Fulu(Hc::from(apply_fulu_delta((**base).clone(), delta)))
+        }
+        _ => panic!("Cannot apply delta from different phase"),
+    }
+}
--- a/fork_choice_control/src/delta/mod.rs
+++ b/fork_choice_control/src/delta/mod.rs
@@ -95,35 +95,29 @@ pub fn delta<P: Preset>(
     }
 }
 
-pub fn apply_delta<P: Preset>(
-    base: &BeaconState<P>,
-    delta: &BeaconStateDelta<P>,
-) -> BeaconState<P> {
+pub fn apply_delta<P: Preset>(base: &BeaconState<P>, delta: BeaconStateDelta<P>) -> BeaconState<P> {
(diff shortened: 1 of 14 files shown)


fork_choice_control/src/storage.rs
```diff
@@ -670,7 +757,53 @@ impl<P: Preset> Storage<P> {
...
+        let Some(delta_bytes) = self.database.get(delta_key)? else {
+            return Ok(None);
+        };
+        let delta: BeaconStateDelta<P> =
+            bincode::deserialize(&delta_bytes).context("failed to deserialize delta")?;
+
+        let slot = delta.slot();
+        let epoch = Self::epoch_at_slot(slot);
+
+        let base_epoch = (epoch / DEFAULT_STATE_ARCHIVAL_EPOCH_INTERVAL.get())
```
Tumas: This value is computed differently [here](https://github.com/galadd/grandine/blob/feature/disk-usage-optimization/fork_choice_control/src/storage.rs#L385), where it is derived from self’s properties rather than global default constants. This logic should be made uniform by extracting it into a shared helper function.

fork_choice_control/src/storage.rs
```diff
@@ -1177,6 +1312,14 @@ impl PrefixableKey for StateByBlockRoot {
     const PREFIX: &'static str = "s";
 }
 
+#[derive(Display)]
+#[display("{}{_0:x}", Self::PREFIX)]
+pub struct StateDeltaByBlockRoot(pub H256);
```
Tumas, line 1336: [db_stats utility command](https://github.com/galadd/grandine/blob/feature/disk-usage-optimization/runtime/src/db_stats.rs#L56) should be updated as well

fork_choice_control/src/delta/phase0.rs
```diff
@@ -0,0 +1,580 @@
...
+    base_slot: Slot,
+    target_slot: Slot,
+    target_roots: &RecentRoots<P>,
+) -> Vec<H256> {
+    let inputs = target_slot - base_slot;
+
+    let roots: Vec<_> = (0..inputs)
+        .map(|i| {
+            *target_roots
+                .get((base_slot + i) % 8192)
```
Tumas: Magic constant: use `SlotsPerHistoricalRoot::<P>::U64`

fork_choice_control/src/delta/phase0.rs
```diff
@@ -0,0 +1,580 @@
...
+    roots
+}
+
+pub(super) fn apply_roots_delta<P: Preset>(
+    base_slot: Slot,
+    mut base_roots: RecentRoots<P>,
+    delta: &[H256],
+) -> RecentRoots<P> {
+    for (i, change) in delta.iter().enumerate() {
+        *base_roots.mod_index_mut((base_slot + i as u64) % 8192) = *change;
```
Tumas: Magic constant: use `SlotsPerHistoricalRoot::<P>::U64`

fork_choice_control/src/delta/phase0.rs
```diff
@@ -0,0 +1,580 @@
...
+    }
+
+    base_roots
+}
+
+pub(super) fn validators_delta<P: Preset>(
+    base_validators: &Validators<P>,
+    target_validators: &Validators<P>,
+) -> ValidatorsChange {
+    let len1 = base_validators.len_usize();
```
Tumas: The validators registry is append-only across finalized states. This property can be leveraged to significantly simplify the computation of validator deltas.

fork_choice_control/src/delta/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        (BeaconState::Fulu(base), BeaconState::Fulu(target)) => {
+            BeaconStateDelta::Fulu(fulu_delta(base, (**target).clone()))
+        }
+        _ => panic!("Cannot create delta across different phases"),
+    }
+}
+
+pub fn apply_delta<P: Preset>(
+    base: &BeaconState<P>,
+    delta: &BeaconStateDelta<P>,
```
Tumas: `delta` argument could be moved into a function instead of being passed as a reference. This would remove the need for all of the extra cloning.

fork_choice_control/src/delta/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        (BeaconState::Deneb(base), BeaconState::Deneb(target)) => {
+            BeaconStateDelta::Deneb(deneb_delta(base, (**target).clone()))
+        }
+        (BeaconState::Electra(base), BeaconState::Electra(target)) => {
+            BeaconStateDelta::Electra(electra_delta(base, (**target).clone()))
+        }
+        (BeaconState::Fulu(base), BeaconState::Fulu(target)) => {
+            BeaconStateDelta::Fulu(fulu_delta(base, (**target).clone()))
+        }
+        _ => panic!("Cannot create delta across different phases"),
```
Tumas: With current logic of persistence of finalized states (epoch matching DEFAULT_STATE_ARCHIVAL_EPOCH_INTERVAL or state_archival_epoch_interval) it is likely that sooner or later delta will be calculated across forks. To resolve this:
- This should return an error instead of panic and the error should be handled appropriately when calling this function.
- There should be logic that prevents delta calculations across forks. For example, full state should be stored in the database if it's a new fork instead of computing deltas.

Similarly, this also applies to applying delta across forks.

Tumas (changes requested):

fork_choice_control/src/storage.rs
```diff
@@ -1177,6 +1326,14 @@ impl PrefixableKey for StateByBlockRoot {
...
+
+impl PrefixableKey for StateDeltaByBlockRoot {
+    const PREFIX: &'static str = "s_d";
```
Tumas: The prefix should be unique for state deltas. It currently includes prefix for `StateByBlockRoot`. The general rule is that thay should be unique. `UnfinalizedBlockByRoot` and `FinalizedBlockByRoot` is an exception to this though.

fork_choice_control/src/storage.rs
```diff
@@ -670,7 +767,52 @@ impl<P: Preset> Storage<P> {
...
+
+        let base_root = self
+            .block_root_by_slot(base_slot)?
+            .ok_or(Error::DependentRootLookupFailed)?;
+
+        if let Ok(cache) = self.base_state_cache.read()
+            && let Some((cached_root, cached_state)) = cache.as_ref()
+            && *cached_root == base_root
+        {
+            let reconstructed = apply_delta(&cached_state.as_ref().clone(), delta)?;
```
Tumas: cached_state is a `&Arc<BeaconState<P>>` here so you can pass it directly

fork_choice_control/src/storage.rs
```diff
@@ -670,7 +767,52 @@ impl<P: Preset> Storage<P> {
...
+        }
+
+        let base_state = self
+            .get::<BeaconState<P>>(StateByBlockRoot(base_root))?
+            .ok_or(Error::StateNotFound {
+                state_slot: base_slot,
+            })?;
+
+        if let Ok(mut cache) = self.base_state_cache.write() {
+            *cache = Some((base_root, base_state.clone().into()));
```
Tumas: If `base_state` were converted to an Arc earlier, you could avoid cloning the underlying state here. Instead, you would clone the Arc itself using `ArcExt::clone_arc`, which is significantly cheaper.

fork_choice_control/src/storage.rs
```diff
@@ -329,27 +342,108 @@ impl<P: Preset> Storage<P> {
...
+
+        if append_delta {
+            info_with_peers!("saving state delta in slot {state_slot}");
+            let base_epoch = self.base_epoch_at_state_epoch(state_epoch);
+            let base_slot = misc::compute_start_slot_at_epoch::<P>(base_epoch);
+            let base_root = self.block_root_by_slot(base_slot)?;
+
+            let base_state = if let Some(base_root_value) = base_root {
+                let cached_state = if let Ok(cache) = self.base_state_cache.read() {
+                    if let Some((cached_root, cached_state)) = cache.as_ref() {
```
Tumas: The current code looks messy.  Consider encapsulating access to `BaseStateCache` by implementing getter and setter methods directly on the `BaseStateCache` type. Or alternatively add helper functions since it's currently a type alias.

fork_choice_control/src/storage.rs
galadd: Can it be like "sd", or it needs to be a single character? "d" is already used. is "x" intuitive

Edit: Just saw that SlotByStateRoot and BlobSidecarByBlobId's prefixes aren't intuitive, so "x" should be fine

fork_choice_control/src/delta/phase0.rs
```diff
@@ -0,0 +1,598 @@
...
+}
+
+pub(super) fn randao_delta<P: Preset>(
+    base_slot: Slot,
+    target_slot: Slot,
+    target_randao: &RandaoMixes<P>,
+) -> RandaoChange {
+    let diff_len = (target_slot - base_slot) / P::SlotsPerEpoch::U64;
+    let end_idx = (target_slot / P::SlotsPerEpoch::U64) % P::EpochsPerHistoricalVector::U64;
+    let start_idx = end_idx - diff_len - 1;
```
Tumas: This will panic when calculating delta between genesis state and the next state after. Use `u64::saturating_sub`

galadd: Actually, it wraps around. My logic doesn't cater for that when creating the delta (error from me). But u64::saturating_sub can't work because it results to 0. So more like,

`let start_idx = (end_idx + P::EpochsPerHistoricalVector::U64 - diff_len - 1) % P::EpochsPerHistoricalVector::U64;`

and then use `mod_index`, as opposed to `get` in creating delta

galadd: applies to every other vector

fork_choice_control/src/storage.rs
```diff
@@ -670,7 +749,54 @@ impl<P: Preset> Storage<P> {
...
+        let state_epoch = Self::epoch_at_slot(slot);
+
+        let base_epoch = self.base_epoch_at_state_epoch(state_epoch);
+        let base_slot = misc::compute_start_slot_at_epoch::<P>(base_epoch);
+
+        let base_root = self
+            .block_root_by_slot(base_slot)?
+            .ok_or(Error::DependentRootLookupFailed)?;
+
+        if let Ok(cache) = self.base_state_cache.read()
```
Tumas: Newly  added helper method `Storage::load_cached_base_state` could be used here

fork_choice_control/src/storage.rs
```diff
@@ -670,7 +749,54 @@ impl<P: Preset> Storage<P> {
...
+
+        let base_state = self
+            .get::<BeaconState<P>>(StateByBlockRoot(base_root))?
+            .ok_or(Error::StateNotFound {
+                state_slot: base_slot,
+            })?;
+
+        let base_state_arc = Arc::new(base_state);
+
+        if let Ok(mut cache) = self.base_state_cache.write() {
```
Tumas: Newly added helper method `Storage::set_cached_base_state` could be used here

fork_choice_control/src/storage.rs
```diff
@@ -352,6 +354,80 @@ impl<P: Preset> Storage<P> {
...
+            && state_epoch.is_multiple_of(self.state_archival_epoch_interval.into());
+        let append_delta = misc::is_epoch_start::<P>(state_slot)
+            && state_epoch.is_multiple_of(self.archival_epoch_interval.into())
+            && !append_state;
+
+        if append_state {
+            info_with_peers!("saving state in slot {state_slot}");
+            let current_state = state.get_or_init(|| chain_link.state(store));
+            batch.push(serialize(StateByBlockRoot(block_root), current_state)?);
+            if let Ok(mut base_state) = self.base_state_cache.write() {
```
Tumas: Newly added helper method `Storage::set_cached_base_state` could be used here

fork_choice_control/src/storage.rs
```diff
@@ -806,11 +806,13 @@ impl<P: Preset> Storage<P> {
                 state_slot: base_slot,
             })?;
 
+        let base_state_arc = Arc::new(base_state);
+
         if let Ok(mut cache) = self.base_state_cache.write() {
-            *cache = Some((base_root, base_state.clone().into()));
+            *cache = Some((base_root, Arc::clone(&base_state_arc)));
```
Tumas: It's more conventional in this project to use `ArcExt::clone_arc` here. See other uses of `clone_arc` in the project

fork_choice_control/src/delta/phase0.rs
```diff
@@ -0,0 +1,590 @@
...
+pub(super) fn randao_delta<P: Preset>(
+    base_slot: Slot,
+    target_slot: Slot,
+    target_randao: &RandaoMixes<P>,
+) -> RandaoChange {
+    let modulus = P::EpochsPerHistoricalVector::U64;
+
+    let diff_len = (target_slot - base_slot) / P::SlotsPerEpoch::U64;
+    let end_idx = (target_slot / P::SlotsPerEpoch::U64) % modulus;
+    let start_idx = (end_idx + modulus - diff_len - 1) % modulus;
```
Tumas: [This commit](https://github.com/grandinetech/grandine/pull/523/changes/ccfe2abc840436165af542649d8bd5297b0b1798) seems to have introduced a regressesion. You can find it by adding and running some sanity tests for your code:

```
   //fork_choice_control/src/storage.rs: 

    #[cfg(feature = "eth2-cache")]
    #[test_case(
        Config::mainnet().into(),
        eth2_cache_utils::mainnet::GENESIS_BEACON_STATE.force(),
        eth2_cache_utils::mainnet::BEACON_BLOCKS_UP_TO_SLOT_128.force()
    )]
    // #[test_case(
    //     Config::mainnet().into(),
    //     eth2_cache_utils::mainnet::ALTAIR_BEACON_STATE.force(),
    //     eth2_cache_utils::mainnet::ALTAIR_BEACON_BLOCKS_FROM_128_SLOTS.force()
    // )]
    // #[test_case(
    //     Config::mainnet().into(),
    //     eth2_cache_utils::mainnet::CAPELLA_BEACON_STATE.force(),
    //     eth2_cache_utils::mainnet::CAPELLA_BEACON_BLOCKS_FROM_244816_SLOTS.force()
    // )]
    fn test_deltas_roundtrip<P: Preset>(
        config: Arc<Config>,
        base_state: &Arc<BeaconState<P>>,
        blocks: &[Arc<SignedBeaconBlock<P>>],
    ) -> Result<()> {
        let pubkey_cache = Arc::new(PubkeyCache::default());

        let mut state = base_state.clone_arc();
        let state = state.make_mut();

        for block in blocks.iter().skip(1) {
            combined::untrusted_state_transition(&config, &pubkey_cache, state, block)?;

            let delta = delta(&base_state, &Arc::new(state.clone()))?;
            let recovered = apply_delta(&base_state, delta)?;

            assert_eq!(*state, recovered);
        }

        Ok(())
    }
```

To run these tests modify `make tests` command in the `Makefile`:
```
cargo test --release --features stub-grandine-version $(FEATURES) --features eth2-cache $(EXCLUDES) -p fork_choice_control delta -- --nocapture
```

and clone the https://github.com/grandinetech/eth2-cache to your local Grandine's project root

PR #529, opened 2025-12-10
Replace `dashmap` with `scc`

Tumas:


benches/benches/fork_choice_store.rs
```diff
@@ -86,7 +86,7 @@ impl Criterion {
                     false,
                     false,
                     [].into(),
-                    Arc::new(DashMap::new()),
+                    Arc::new(HashMap::new()),
```
hangleang: Can we alias the type to avoid confusion with std lib `HashMap`?

povi (approved):

PR #533, opened 2025-12-11
Return `PhaseError` on block/state phase mismatch when processing blocks

Tumas:


transition_functions/src/combined.rs
```diff
@@ -212,12 +212,15 @@ pub fn custom_state_transition<P: Preset>(
...
             slot_report,
         ),
-        _ => {
+        (state, block) => {
             // This match arm will silently match any new phases.
             // Cause a compilation error if a new phase is added.
             const_assert_eq!(Phase::CARDINALITY, 7);
 
-            unreachable!("successful slot processing ensures that phases match")
+            bail!(PhaseError {
```
povi, line 221: Give more info (at least add block root and slot, and state slot to the message)

povi (approved):

PR #537, opened 2025-12-16
Handle block payload attestations

hangleang:


fork_choice_control/src/misc.rs
```diff
@@ -162,11 +163,7 @@ pub struct PendingAggregateAndProof<P: Preset> {
 
 pub type PendingAttestation<P> = AttestationItem<P, GossipId>;
 
-#[derive(Debug)]
-pub struct PendingPayloadAttestation {
-    pub payload_attestation: Arc<PayloadAttestationMessage>,
-    pub origin: PayloadAttestationOrigin,
-}
+pub type PendingPayloadAttestation<P> = PayloadAttestationItem<P>;
```
povi: Remove type alias, and just use `PayloadAttestationItem`. Less types - less confusion

povi (approved):

PR #538, opened 2025-12-17
Delay execution payload envelope until data available

hangleang:


fork_choice_control/src/mutator.rs
```diff
@@ -1709,6 +1711,8 @@ where
                     } else {
                         self.send_to_p2p(P2pMessage::BlockNeeded(block_root, peer_id));
                     }
+
+                    // TODO: (gloas): retry payload envelope corresponding to `block_root`?
```
povi: Here? No

povi (approved):

fork_choice_control/src/mutator.rs
hangleang: I'll remove the TODO

PR #548, opened 2026-01-02
Update zkvms

ArtiomTr:


zkvm/guest/risc0/guest/src/main.rs
```diff
@@ -54,20 +56,20 @@ fn read_block_and_state<P: Preset>() -> Result<(Config, SignedBeaconBlock<P>, Be
     Ok((config, block, state, cache))
 }
 
-fn main() -> Result<()> {
+fn main() {
```
povi, line 59: What's the reason for this change?

ArtiomTr: risc0 v3 has new entrypoint declaration procedure, which doesn't support `Result` in main method return

povi (approved):

PR #550, opened 2026-01-05
Add back-sync support for execution payload envelope

Subhasish-Behera:


p2p/src/back_sync.rs
```diff
@@ -435,6 +436,56 @@ impl<P: Preset> Batch<P> {
             .collect()
     }
 
+    fn valid_execution_payload_envelopes_for(
+        &self,
+        config: &Config,
+        controller: &RealController<P>,
+        block: &Arc<SignedBeaconBlock<P>>,
+        parent: &Arc<SignedBeaconBlock<P>>,
```
hangleang: I don't think `parent_block` is needed

transition_functions/src/lib.rs
```diff
@@ -238,6 +238,9 @@ pub mod fulu {
 
 pub mod gloas {
     pub use block_processing::get_expected_withdrawals;
+    pub use execution_payload_processing::{
+        process_execution_payload, verify_execution_payload_envelope_signature,
```
hangleang: export without use

hangleang (commented):
Look good, except few nits

p2p/src/back_sync.rs
```diff
@@ -435,6 +436,56 @@ impl<P: Preset> Batch<P> {
...
+        }
+
+        debug_with_peers!(
+            "validated execution payload envelope for back sync (block_root: {:?}, slot: {}, parent_slot: {})",
+            block_root,
+            slot,
+            parent.message().slot()
+        );
+
+        Ok(vec![envelope.clone_arc()])
```
hangleang: this could return a single item, there won't be more than 1 payload per block

hangleang (approved):

PR #555, opened 2026-01-05
Disable check for --eth1-rpc-urls in embedded mode

ArtiomTr:
Disabled check for --eth1-rpc-urls being present, when grandine is compiled with "embed" feature. There is no need for this flag to be specified, as all engine API communication happens through FFI calls.


Tumas (commented):
This should probably be updated too: 

https://github.com/grandinetech/grandine/blob/develop/runtime/src/runtime.rs#L362

Tumas (approved):

PR #557, opened 2026-01-05
Update rust toolchain to 1.92.0

ArtiomTr:
Update rust to latest stable toolchain version. This also updates cross to latest version.


Tumas (commented):
I think the exception: 

```
#![expect(unexpected_cfgs, reason = "construct_fixed_hash! uses cfg internally")]
```

can be removed from:

https://github.com/grandinetech/grandine/blob/develop/bls/bls-blst/src/public_key_bytes.rs#L1

and 

https://github.com/grandinetech/grandine/blob/develop/bls/bls-blst/src/signature_bytes.rs#L1

and references to Rust version should probably be updated as well:

https://github.com/grandinetech/grandine/blob/develop/arithmetic/src/lib.rs#L6
https://github.com/grandinetech/grandine/blob/develop/benches/benches/hashing.rs#L94
https://github.com/grandinetech/grandine/blob/develop/http_api_utils/src/traits.rs#L15
https://github.com/grandinetech/grandine/blob/develop/slasher/src/status.rs#L5
https://github.com/grandinetech/grandine/blob/develop/ssz/src/merkle_tree.rs#L247

povi (approved):

PR #576, opened 2026-01-28
Improve own committee members calculation

Tumas:
- keep results scoped under a dependent root
- move precalculations and pruning tasks to a dedicated executor


validator/src/validator.rs
```diff
@@ -2047,56 +2061,40 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
...
-                {
-                    update_beacon_committee_subscriptions(
-                        current_slot,
-                        &members,
-                        &subnet_service_tx,
-                    )
-                    .await;
-                }
-            }
+        let task = UpdateBeaconCommitteeSubscriptionsTask {
```
povi, line 2086: In the future, if we move all validator parts with signing to tasks, we should remove `spawn_blocking` from signer.

validator/src/tasks.rs
```diff
@@ -0,0 +1,121 @@
...
+            controller,
+            mut beacon_state,
+            own_beacon_committee_members,
+            subnet_service_tx,
+            wait_group,
+        } = self;
+
+        let current_slot = beacon_state.slot();
+
+        for slot in OwnBeaconCommitteeMembers::slots_to_compute_in_advance(current_slot) {
```
povi, line 43: dependent root for n - 1 epoch is not enough to make sure committees in n + 1 epoch will not change

validator/src/tasks.rs
```diff
@@ -0,0 +1,121 @@
...
+                            warn_with_peers!(
+                                "failed to preprocess next fork beacon state for beacon committee \
+                                subscriptions: {error:?}",
+                            );
+                            break;
+                        }
+                    }
+                }
+
+                info_with_peers!("updating beacon committee subscriptions {slot} {current_slot}");
```
povi: debug level

povi (approved):

PR #579, opened 2026-01-29
feat: Add ProposerPreferences gossip handling and broadcasting

Subhasish-Behera:
Implement end-to-end support for ProposerPreferences gossip topic, allowing validators to broadcast their fee_recipient and gas_limit preferences for upcoming proposal slots in the next epoch.

- Add DOMAIN_PROPOSER_PREFERENCES constant and SignForSingleFork impl
- Add is_valid_proposal_slot and get_upcoming_proposal_slots accessors
- Implement gossip receiving
- Add validate_proposer_preferences for validation
- add bid validation conditions
- Add batch signing in validator
- Store preferences in fork_choice_store for bid selection integration


fork_choice_store/src/store.rs
```diff
@@ -1888,6 +1889,82 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+        // (stale state's proposer_lookahead may not cover next epoch from tick clock).
+        //
+        // existing_state_at_slot returns None if validator tick (through preprocessed_state_post_block_blocking)
+        // hasn't advanced the state to the current slot yet (gossip validation is(seems) decoupled from tick processing).
+        // In that case we Ignore — the message. so small gap of gossip arriving before tick can give None.
+        // if manual process_slots is not a option + if any DelayUntil* doesnt fit the situation, then a fallback case to call with (head_root, N-1)
+        // on same side of epoch might be option.
+        let Some(state) =
+            self.state_cache
+                .existing_state_at_slot(self, self.head().block_root, self.slot())
```
povi: You can use `StateCache::state_at_slot` method, which will do slot processing if there is a state in state cache with given block root, but not in given slot. That should resolve the concerns you expressed in the comment.

Subhasish-Behera: fixed

p2p/src/network.rs
hangleang: It is supposed to subscribe once at the start of the epoch

Subhasish-Behera: hey @hangleang e9b96eea89ad4e65e236b088c042d25e515a2135 should fix this problem. I am reusing `last_nfd_update_epoch` field to use it as `last_visited_epoch`. If you prefer a separate field(like `last_proposer_preferences_epoch`) to address this particular problem, which is not really required, we can do that as well(let me know if thats the case). that would be similar to what is done in Validator's (`last_*_epoch` fields) struct in `validator/src/validator.rs`. but there cgc udpate and proposer pref division was handled separately in comparison to network.rs, though the areas impacting is same

PR #580, opened 2026-01-29
Use client versions in builder blocks

Tumas:


eth1_api/src/eth1_api/embed_api.rs
```diff
@@ -160,6 +160,10 @@ impl Eth1Api {
         Self { config, metrics }
     }
 
+    pub fn client_versions(&self) -> impl Iterator<Item = Arc<ClientVersions>> {
+        core::iter::empty()
```
povi, line 165: Please add a `TODO` here if this is out of scope

povi (approved):

PR #589, opened 2026-02-04
Update consensus-spec-tests to v1.7.0-alpha.2

hangleang:
Reference to spec release: https://github.com/ethereum/consensus-specs/releases/tag/v1.7.0-alpha.2

### DIFF
--- a/transition_functions/src/gloas/block_processing.rs
+++ b/transition_functions/src/gloas/block_processing.rs
@@ -24,7 +24,7 @@ use helper_functions::{
         balance, builder_balance, decrease_balance, increase_balance, initiate_builder_exit,
     },
     predicates::{
-        can_builder_cover_bid, is_active_builder, is_attestation_same_slot, is_parent_block_full,
+        can_builder_cover_bid, is_active_builder, is_attestation_same_slot,
         validate_constructed_indexed_attestation, validate_constructed_indexed_payload_attestation,
     },
     signing::SignForSingleFork as _,
@@ -153,6 +153,7 @@ pub fn custom_process_block<P: Preset>(
     process_withdrawals(state)?;
 
     // > [New in Gloas:EIP7732]
+    // This function must be called after `process_withdrawals`
     process_execution_payload_bid(config, pubkey_cache, state, block)?;
 
     unphased::process_randao(config, pubkey_cache, state, &block.body, &mut verifier)?;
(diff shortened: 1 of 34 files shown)


transition_functions/src/gloas/block_processing.rs
```diff
@@ -432,7 +444,8 @@ fn process_validators_sweep_withdrawals<P: Preset>(
 }
 
 pub fn process_withdrawals<P: Preset>(state: &mut impl PostGloasBeaconState<P>) -> Result<()> {
-    if !is_parent_block_full(state) {
```
povi, line 435: Why it was inlined and the helper removed?

povi: If there is no serious reason to do otherwise, it better to have helper and other methods as it is in specs. It makes code search / updates / etc. easier

hangleang: it's only used once, and the function is shallow, so better to be inlined

hangleang: it is just for convenient reason, no strongly opinion about it

transition_functions/src/gloas/block_processing.rs
```diff
@@ -226,11 +235,12 @@ fn get_builder_withdrawals_count<P: Preset>(
     state: &(impl PostGloasBeaconState<P> + ?Sized),
     withdrawal_index: &mut WithdrawalIndex,
     withdrawals: &mut Vec<Withdrawal>,
+    withdrawal_limit: usize,
 ) -> Result<usize> {
     let mut processed_count = 0;
 
     for withdrawal in &state.builder_pending_withdrawals().clone() {
-        if withdrawals.len() == P::MaxWithdrawalsPerPayload::USIZE {
+        if withdrawals.len() == withdrawal_limit {
```
povi: This check should be `withdrawals.len() >= withdrawal_limit` here and everywhere else where withdrawals len is measured against limit. Please refer to the spec: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/beacon-chain.md#new-get_builders_sweep_withdrawals

povi: This applies to other phases too. See https://github.com/ethereum/consensus-specs/pull/4832

hangleang: I haven't refactor [`get_expected_withdrawals`](https://github.com/ethereum/consensus-specs/pull/4766) function for older phases, and the expected `withdrawal_limit` is passed to those functions so I assume keeping the check as it is fine, but I'm also okay with the change to `>=` to match the spec, or unless backport to old phases is needed if so full refactor.

povi (approved):
LGTM

PR #593, opened 2026-02-06
Preload first and the last block of previous epoch into `state_at_slot_cache` when archiving finalized blocks

Tumas:


fork_choice_control/src/mutator.rs
```diff
@@ -3464,6 +3469,37 @@ where
                         Ok(slots) => {
                             if let Some(chain_link) = archived.back() {
                                 let finalized_block = chain_link.block.clone_arc();
+                                let anchor_epoch = misc::compute_epoch_at_slot::<P>(chain_link.slot());
+
+                                for candidate in &archived {
+                                    let candidate_slot = candidate.slot();
+                                    let candidate_epoch = misc::compute_epoch_at_slot::<P>(candidate_slot);
+
+                                    let is_last_block = candidate_slot + 1 == chain_link.slot();
```
povi, line 3478: What if the slot is empty?

povi (approved):

fork_choice_control/src/mutator.rs
Tumas: state is not preloaded to state_at_slot_cache, upon request defaults to retrieval from storage and block processing

PR #596, opened 2026-02-08
Moved column (re)construction to dedicated executor

ArtiomTr:
Both `compute_cells` & `recover_cells_and_kzg_proofs` were moved from a rayon threadpool to a dedicated executor. This change, while introducing unnecessary cloning of input data (due to dedicated executor lifetime limitations), reduces p99 latency during high load.

Cell computation is prioritized over cell recovery, as it requires less computation resources.


eip_7594/src/lib.rs
```diff
@@ -245,9 +243,10 @@ pub fn verify_sidecar_inclusion_proof<P: Preset>(
     )
 }
 
-pub fn recover_matrix<P: Preset>(
+pub async fn recover_matrix<P: Preset>(
     partial_matrix: &[MatrixEntry<P>],
```
Tumas: You could change `partial_matrix` argument to a Vec and move argument into a function. This would allow to remove extra cloning in `e.cell.clone()`. 

Having a reference argument doesn't provide any benefits as all calls to `recover_matrix` no longer use partial_matrix argument after calling it.

eip_7594/src/lib.rs
```diff
@@ -407,10 +416,11 @@ pub fn construct_data_column_sidecars_from_sidecar<P: Preset>(
     }
 }
 
-pub fn try_convert_to_cells_and_kzg_proofs<P: Preset>(
+pub async fn try_convert_to_cells_and_kzg_proofs<P: Preset>(
     blobs: &[Blob<P>],
```
Tumas: Same here. Move blobs argument into a function by changing it into: 
```rust
    blobs: impl ExactSizeIterator<Item = Blob<P>>,
```

this would allow `blobs.len()` in the function and would eliminate need for blob cloning.

Then also change blob iteration inside a function into `blobs.into_iter()`

And finally change blobs argument when calling this function into `blobs.into_iter()`

Tumas (changes requested):

eip_7594/src/lib.rs
ArtiomTr: good catch, fixed this :+1:

eip_7594/src/lib.rs
ArtiomTr: fixed :+1:

operation_pools/src/blob_reconstruction_pool/tasks.rs
```diff
@@ -82,24 +86,27 @@ impl<P: Preset, W: Wait> PoolTask for ReconstructDataColumnSidecarsTask<P, W> {
...
-            eip_7594::recover_matrix(&partial_matrix, controller.store_config().kzg_backend)
-                .tap(|_| prometheus_metrics::stop_and_record(columns_reconstruction_timer))
-                .and_then(|full_matrix| {
-                    let timer = metrics
-                        .as_ref()
-                        .map(|metrics| metrics.data_column_sidecar_computation.start_timer());
+        let reconstruction_result = eip_7594::recover_matrix(
+            partial_matrix,
+            controller.store_config().kzg_backend,
+            dedicated_executor.clone_arc(),
```
Tumas: unnecessary `clone_arc` here, `dedicated_executor` is no longer used after this

ArtiomTr: fixed this

Tumas (approved):

PR #597, opened 2026-02-09
`GET /eth/v2/validator/duties/proposer/{epoch}`

hangleang:
Reference spec: https://github.com/ethereum/beacon-APIs/commit/9649bf35d461f3f8cf320852dde54d55bb44797a

### DIFF
--- a/http_api/src/standard.rs
+++ b/http_api/src/standard.rs
@@ -2530,6 +2530,42 @@ pub async fn validator_proposer_duties<P: Preset, W: Wait>(
     State(anchor_checkpoint_provider): State<AnchorCheckpointProvider<P>>,
     EthPath(epoch): EthPath<Epoch>,
 ) -> Result<EthResponse<Vec<ValidatorProposerDutyResponse>>, Error> {
+    get_proposer_duties(
+        chain_config,
+        controller,
+        anchor_checkpoint_provider,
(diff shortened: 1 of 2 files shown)


http_api/src/standard.rs
```diff
@@ -2579,6 +2579,73 @@ pub async fn validator_proposer_duties<P: Preset, W: Wait>(
         .execution_optimistic(status.is_optimistic()))
 }
 
+// This endpoint is introduced to account for deterministic proposer lookahead
+// and return a different `dependent_root` based on current fork.
+/// `GET /eth/v2/validator/duties/proposer/{epoch}`
+pub async fn validator_proposer_duties_v2<P: Preset, W: Wait>(
```
Tumas, line 2546: This is pretty much the same as V1 version, except for dependent root value. Extracting duplicated code into a helper method would make this fact obvious.

Tumas (changes requested):

http_api/src/standard.rs
hangleang: Changed at https://github.com/grandinetech/grandine/pull/597/changes/172d930de70bf76d96d3abcfdf49fffbc47ed411

Tumas (approved):

PR #599, opened 2026-02-10
`GET /eth/v2/node/version`

hangleang:
Implement https://github.com/ethereum/beacon-APIs/pull/568

### DIFF
--- a/http_api/src/standard.rs
+++ b/http_api/src/standard.rs
@@ -23,7 +23,7 @@ use block_producer::{BlockBuildOptions, BlockProducer, ProposerData, ValidatorBl
 use bls::{PublicKeyBytes, SignatureBytes, traits::SignatureBytes as _};
 use builder_api::unphased::containers::SignedValidatorRegistrationV1;
 use enum_iterator::Sequence as _;
-use eth1_api::{ApiController, Eth1Api};
+use eth1_api::{ApiController, ClientVersionV1, Eth1Api};
 use eth2_libp2p::{GossipId, PeerId};
 use fork_choice_control::{Event, EventChannels, ForkChoiceContext, ForkTip, Topic, Wait};
 use fork_choice_store::{AttestationItem, AttestationOrigin};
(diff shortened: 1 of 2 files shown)


http_api/src/standard.rs
```diff
@@ -2406,6 +2413,19 @@ pub async fn node_version(State(network_config): State<Arc<NetworkConfig>>) -> R
...
 }
 
+/// `GET /eth/v2/node/version`
+pub async fn node_version_v2(State(eth1_api): State<Arc<Eth1Api>>) -> Response {
+    let data = NodeVersionV2Response {
+        beacon_node: ClientVersionV1::own(),
+        execution_client: eth1_api
+            .client_versions()
+            .next()
+            .and_then(|versions| versions.first().cloned()),
```
Tumas, line 2423: If `execution_client` field is changed to an array:

```diff
-        execution_client: eth1_api
-            .client_versions()
-            .next()
-            .and_then(|versions| versions.first().cloned()),
+        execution_client: eth1_api.client_versions().next(),

```

This would use the client_versions of the first specified EL (it returns an array if it is a mulitplexer). It is not correct in case there are multiple ELs attached to Grandine (by specifying multiple `--eth1-rpc-urls` options) and the first one being down, while the second one being used. This edge case could be fixed by taking the client versions of the "active" attached EL. 

The "active" EL is determined by calling 'Endpoints::endpoints_for_request' and taking the first returned option.

http_api/src/standard.rs
```diff
@@ -411,6 +411,13 @@ struct NodeVersionResponse<'version> {
     version: Option<&'version str>,
 }
 
+#[derive(Serialize)]
+struct NodeVersionV2Response {
+    beacon_node: ClientVersionV1,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    execution_client: Option<ClientVersionV1>,
```
Tumas, line 418: This field is supposed to be an array:

https://github.com/ethereum/beacon-APIs/pull/568/changes#diff-681f651b832677b3dabee0549130a42382aa47297b33092ad3d369d084533828R32

```diff
-    execution_client: Option<ClientVersionV1>,
+    execution_client: Option<Arc<ClientVersions>>,
```

Tumas (changes requested):

http_api/src/standard.rs
hangleang: Yes, it just changed after I made the PR. I will wait until the spec is merged

nflaig: fyi, we've decided to only return the first value again and not an array

Tumas (approved):

PR #600, opened 2026-02-10
Update `POST /eth/v2/beacon/blocks` to publish only signed beacon block for post-Gloas

hangleang:


Tumas (approved):

Tumas (changes requested):
```SignedAPIBlock::Gloas(SignedGloasBlockWithBlobs<P>)``` is not a correct definition. There are no `blobs` and `kzg_proofs` fields. This could be updated with this PR.

PR #602, opened 2026-02-16
Check pubkey in pending deposits before register builder

hangleang:
Reference to these two spec changes: https://github.com/ethereum/consensus-specs/commit/b3341d00639c4b9c58e02620404663bcca954ab3 and https://github.com/ethereum/consensus-specs/commit/40504e4c431baa88fb8a6b694286b13594fee6da

### DIFF
--- a/transition_functions/src/gloas/execution_payload_processing.rs
+++ b/transition_functions/src/gloas/execution_payload_processing.rs
@@ -1,4 +1,5 @@
 use anyhow::{Result, ensure};
+use bls::PublicKeyBytes;
 use execution_engine::ExecutionEngine;
 use helper_functions::{
     error::SignatureKind,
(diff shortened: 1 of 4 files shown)


transition_functions/src/gloas/execution_payload_processing.rs
```diff
@@ -324,6 +340,29 @@ pub fn process_deposit_request<P: Preset>(
...
+    signature_cache: &mut DepositSignatureCache,
+) -> bool {
+    for deposit in state.pending_deposits() {
+        if deposit.pubkey != pubkey {
+            continue;
+        }
+
+        if *signature_cache
+            .entry((pubkey, deposit.slot))
+            .or_insert_with(|| electra::is_valid_deposit_signature(config, pubkey_cache, deposit))
```
Tumas, line 371: `is_valid_deposit_signature` does not use slot for checking signature validity (also amount could be different at the same slot with the same pubkey but invalid signature) so it's probably best to use just `DepositMessage` + signature as a cache key.

Tumas (changes requested):

transition_functions/src/gloas/execution_payload_processing.rs
hangleang: that make sense, I changed it in https://github.com/grandinetech/grandine/pull/602/changes/61de6c9764acfb7f00eae400fa47be6f44bc18b6

https://tenor.com/view/omg-wow-oh-really-makes-sense-gif-17107984

Tumas (approved):

PR #603, opened 2026-02-16
Add payload attestation tick

hangleang:


clock/src/lib.rs
```diff
@@ -258,6 +269,65 @@ pub enum TickKind {
...
+    fn ticks_per_slot<P: Preset>(config: &Config, slot: Slot) -> usize {
+        if config.phase_at_slot::<P>(slot) >= Phase::Gloas {
+            Self::CARDINALITY // 4 intervals of 4 ticks each
+        } else {
+            Self::pre_gloas_ticks().count() // 3 intervals of 4 ticks each
+        }
+    }
+
+    fn pre_gloas_ticks() -> impl Iterator<Item = Self> {
+        [
```
Tumas: this could be implemented without duplicating the list: 

```rust
enum_iterator::all::<Self>().take_while(|t| !matches!(t, Self::PayloadAttest))
```

clock/src/lib.rs
```diff
@@ -269,26 +339,31 @@ pub enum ClockError {
...
+    let (mut next_tick, next_instant) = next_tick_with_instant::<P, _, _>(
+        config,
+        now_instant,
+        now_system_time,
+        genesis_time_in_ms,
+        false,
+    )?;
 
-    let tick_duration = tick_duration(config);
+    let tick_duration = tick_duration_at_slot::<P>(config, next_tick.slot);
```
Tumas: `tick_duration` is computed once at startup (runtime.rs), so if the node starts pre-Gloas, (1s/tick) the interval never adjusts when the Gloas fork occurs (750ms/tick). This case probably warrants a unit-test.

hangleang: address in https://github.com/grandinetech/grandine/pull/603/changes/7f8371a7d6c17de1d737f893d5a4f2614b7d10db

Tumas: It now adjusts to a new tick duration after a fork, but the first tick of the first Gloas slot is emitted prematurely. At the end of the last Fulu tick, it calculates tick duration for the next tick and slot being in a new fork it results in  new duration (which is 750ms) and emits it after this time while still being in Fulu. 

You can verify this in tests:

```diff
         // Fork transition to Gloas at slot 32
-        tokio::time::advance(one_second_duration).await;
+        tokio::time::advance(post_gloas_tick_duration).await;
+        assert_eq!(
+            next_tick(&mut ticks).await?,
+            None,
+            "Propose(32) must not fire 750ms early"
+        );
+
+        tokio::time::advance(Duration::from_millis(250)).await;
         assert_eq!(
             next_tick(&mut ticks).await?,
             Some(Tick::new(32, TickKind::Propose))
         );
```

hangleang: good catch sir! it's fixed at https://github.com/grandinetech/grandine/pull/603/changes/77ec7fbb3ac80e38dd3489ec48b5552c36b3e204

clock/src/lib.rs
Tumas: This is no longer implemented using 'Interval`. This documentation should be updated to reflect sleep_until + try_unfold approach. See also below for more references that are no longer relevant.

clock/src/lib.rs
```diff
@@ -309,23 +378,33 @@ pub fn next_interval_with_remaining_time(
...
     now_instant: I,
     now_system_time: S,
-    genesis_time: UnixSeconds,
+    genesis_time_in_ms: u64,
     only_interval_ticks: bool,
 ) -> Result<(Tick, I)> {
     let unix_epoch_to_now = now_system_time.duration_since(S::UNIX_EPOCH)?;
-    let unix_epoch_to_genesis = Duration::from_secs(genesis_time);
+
+    // Tick is now operate in milliseconds, so to prevent precision loss by rounding down to unix seconds,
```
Tumas: Grammar typo, should be 'Tick now operates'

clock/src/lib.rs
```diff
@@ -269,37 +324,51 @@ pub enum ClockError {
...
+    )?;
+
+    Ok(Box::pin(futures::stream::try_unfold(
+        (next_tick, next_instant.into()),
+        move |(mut next_tick, mut deadline)| {
+            async move {
+                loop {
+                    tokio::time::sleep_until(deadline).await;
+
+                    // Recompute duration based on the *next* tick's slot so the sleep correctly accounts for a Gloas fork crossing.
```
Tumas: Not sure if comment is still correct. Duration is no longer based on the next tick. Here next_tick is a 'current tick'

clock/src/lib.rs
hangleang: updated at https://github.com/grandinetech/grandine/pull/603/changes/5c1e78365839930f5abd795b7aad8f5921231244

PR #611, opened 2026-02-26
Remove validator public keys from persisted state

ArtiomTr:
This change optimizes disk usage, by excluding validator pubkeys from state before persisting it to disk, and restoring them by taking pubkeys from finalized validator list. Pubkeys are removed by zeroizing them, so that snappy compresses them, effectively removing them from state.

This change reduces mainnet state size by ~72% - from ~151.6 to ~41.3 MB compressed size.

### DIFF
--- a/fork_choice_control/src/mutator.rs
+++ b/fork_choice_control/src/mutator.rs
@@ -368,7 +368,7 @@ where
 
         let head_slot = self
             .storage
-            .checkpoint_state_slot()?
+            .checkpoint_state_slot(&self.store.finalized_validators())?
             .unwrap_or_else(|| last_block.message().slot());
 
         self.handle_tick(&wait_group, Tick::start_of_slot(head_slot))?;
@@ -2259,6 +2259,7 @@ where
             let store = self.owned_store();
             let storage = self.storage.clone_arc();
             let wait_group = wait_group.clone();
+            let finalized_validators = store.finalized_validators();
 
             if !unloaded.is_empty() {
                 Builder::new()
@@ -2270,7 +2271,7 @@ where
                         .iter()
                         .map(|chain_link| (chain_link.state(&store), chain_link.block_root));
 
-                    match storage.append_states(states_with_block_roots) {
+                    match storage.append_states(states_with_block_roots, &finalized_validators) {
                         Ok(slots) => {
                             debug_with_peers!(
                                 "unloaded old beacon states persisted \
--- a/fork_choice_control/src/storage.rs
+++ b/fork_choice_control/src/storage.rs
@@ -2,22 +2,25 @@ use core::{cell::OnceCell, marker::PhantomData, num::NonZeroU64};
 use std::{borrow::Cow, sync::Arc};
 
 use anyhow::{Context as _, Error as AnyhowError, Result, bail, ensure};
+use bls::PublicKeyBytes;
 use database::{Database, PrefixableKey};
 use derive_more::Display;
 use fork_choice_store::{ChainLink, Store};
 use genesis::AnchorCheckpointProvider;
 use helper_functions::{accessors, misc};
-use itertools::Itertools as _;
+use itertools::{Either, Itertools as _};
 use logging::{debug_with_peers, info_with_peers, warn_with_peers};
 use nonzero_ext::nonzero;
 use pubkey_cache::PubkeyCache;
 use reqwest::Client;
-use ssz::{Ssz, SszRead, SszReadDefault, SszWrite};
+use ssz::{PersistentList, Ssz, SszRead, SszReadDefault, SszWrite};
 use std_ext::ArcExt as _;
 use thiserror::Error;
 use tracing::info;
 use transition_functions::combined;
+use try_from_iterator::TryFromIterator;
 use types::{
+    Validators,
     combined::{BeaconState, DataColumnSidecar, SignedBeaconBlock},
     config::Config,
     deneb::{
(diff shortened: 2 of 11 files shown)


fork_choice_control/src/mutator.rs
```diff
@@ -368,7 +368,7 @@ where
 
         let head_slot = self
             .storage
-            .checkpoint_state_slot()?
+            .checkpoint_state_slot(self.store.last_finalized().state(&self.store).validators())?
```
povi: `store.last_finalized().state(&self.store).validators()` is repeated too many times.

Please create and use `Store::finalized_state_validators()` method instead (or `finalized_validators` or something).

fork_choice_control/src/storage.rs
```diff
@@ -217,28 +220,40 @@ impl<P: Preset> Storage<P> {
...
         let state_storage = (anchor_state, anchor_block, unfinalized_blocks);
 
         Ok((state_storage, loaded_from_remote))
     }
 
-    fn load_latest_state(&self) -> Result<OptionalStateStorage<'_, P>> {
-        if let Some((state, block, blocks)) = self.load_state_and_blocks_from_checkpoint()? {
+    fn load_latest_state(
+        &self,
+        finalized_validators: Option<&PersistentList<Validator, P::ValidatorRegistryLimit>>,
```
povi: A type alias could be used instead of `PersistentList<Validator, P::ValidatorRegistryLimit>`, as it too is used many times in this file.

Maybe something like `pub type ValidatorList<P> = PersistentList<Validator, P::ValidatorRegistryLimit>` ?

fork_choice_control/src/storage.rs
```diff
@@ -919,6 +1025,63 @@ impl<P: Preset> Storage<P> {
...
+                    )
+                    .context("Invalid finalized validators list")?;
+                validator.pubkey = finalized_validator_pubkey;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn save_validator_pubkeys(
```
povi: Probably better to rename method to `save_finalized_validator_pubkeys` to avoid confusion in the future

fork_choice_control/src/storage.rs
```diff
@@ -1233,6 +1424,15 @@ pub fn serialize(key: impl core::fmt::Display, value: impl SszWrite) -> Result<(
     Ok((serialize_key(key), serialize_value(value)?))
 }
 
+fn prepare_state<P: Preset>(mut state: Arc<BeaconState<P>>) -> Arc<BeaconState<P>> {
+    for validator in &mut *state.make_mut().validators_mut() {
+        // pubkey never changes, so we can restore it later from finalized validator list
+        validator.pubkey = PublicKeyBytes([0; _]);
```
povi: There is a method for that: `PublicKeyBytes::zero()`

fork_choice_control/src/storage.rs
povi: What is more, this method does not save anything by itself. It appends pubkeys to the batch

fork_choice_control/src/storage.rs
```diff
@@ -217,28 +220,40 @@ impl<P: Preset> Storage<P> {
...
             serialize(SlotByStateRoot(anchor_state_root), anchor_slot)?,
-            serialize(StateByBlockRoot(anchor_block_root), &anchor_state)?,
-        ])?;
+            serialize(
+                StateByBlockRoot(anchor_block_root),
+                prepare_state(anchor_state.clone_arc()),
+            )?,
+        ];
+
+        self.save_validator_pubkeys(&mut batch, anchor_state.validators())?;
```
povi: Is this correct? Calling `save_validator_pubkeys` after `prepare_state` would save zero'ed pubkeys to the database

povi: Nvm, there is `clone_arc()`

fork_choice_control/src/mutator.rs
ArtiomTr: fixed this :+1:

fork_choice_control/src/storage.rs
ArtiomTr: replaced with `Validators<P>` from types::collections

fork_choice_control/src/storage.rs
ArtiomTr: renamed to `append_finalized_validator_pubkeys_to_batch`

fork_choice_control/src/storage.rs
ArtiomTr: fixed this :+1:

fork_choice_store/src/misc.rs
```diff
@@ -1076,7 +1077,12 @@ impl DataAvailabilityPolicy {
...
 
-pub trait Storage<P: Preset>: Sync {
+pub trait Storage<P: Preset>: Sync + Sized {
     fn storage_mode(&self) -> StorageMode;
-    fn stored_state_by_block_root(&self, block_root: H256) -> Result<Option<Arc<BeaconState<P>>>>;
+
+    fn stored_state_by_block_root(
+        &self,
+        block_root: H256,
+        finalized_validators: Option<&PersistentList<Validator, P::ValidatorRegistryLimit>>,
```
Tumas: Better to use type alias `Option<&Validators<P>>`

fork_choice_store/src/store.rs
```diff
@@ -3973,7 +3978,14 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
     ) -> Result<Option<Arc<BeaconState<P>>>> {
-        self.storage.stored_state_by_block_root(block_root)
+        let finalized_validators = self
+            .last_finalized()
+            .state
+            .as_ref()
+            .map(types::traits::BeaconState::validators);
+
+        self.storage
+            .stored_state_by_block_root(block_root, finalized_validators)
```
Tumas, line 3998: Is there are reason why this is not `self.storage.stored_state_by_block_root(block_root, Some(&self.finalized_validators()))`? If so, it would be good to have a comment

fork_choice_store/src/misc.rs
ArtiomTr: Fixed this

fork_choice_store/src/store.rs
ArtiomTr: Added comment with explanation above

Tumas (approved):

povi (approved):

PR #628, opened 2026-03-02
Add `--data-availability-window` CLI argument

Tumas:


fork_choice_store/src/store.rs
```diff
@@ -4241,7 +4243,11 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         self.chain_config.deneb_fork_epoch.max(
             self.tick
                 .epoch::<P>()
-                .checked_sub(self.chain_config.min_epochs_for_blob_sidecars_requests)
+                .checked_sub(
+                    self.storage_mode()
```
povi, line 4247: I'd make this a `StorageMode` method that accepts config and does the handling of `custom_data_availability_window` and `unwrap` internally. E.g.:

```
            self.tick
                .epoch::<P>()
                .checked_sub(
                    self.storage_mode()
                          .min_epochs_for_blob_sidecars_requests(&self.chain_config),
               )
```

fork_choice_store/src/store.rs
```diff
@@ -4251,8 +4257,12 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
             self.tick
                 .epoch::<P>()
                 .checked_sub(
-                    self.chain_config
-                        .min_epochs_for_data_column_sidecars_requests,
+                    self.storage_mode()
+                        .custom_data_availability_window()
```
povi: Ditto

povi (approved):

PR #633, opened 2026-03-12
Add `engine_forkchoiceUpdatedV4` support for BAL

hangleang:


ArtiomTr (approved):
C# bindings look good

eth1_api/src/eth1_api/embed_api.rs
```diff
@@ -567,6 +573,30 @@ impl Eth1Api {
...
+                let payload_attributes = payload_attributes
+                    .map(|value| {
+                        if let PayloadAttributes::Gloas(value) = value {
+                            Ok(value)
+                        } else {
+                            Err(Error::InvalidParameters)
+                        }
+                    })
+                    .transpose()?
+                    .clone();
```
Tumas: It appears that extra clone here is unnecessary. The same goes for code for older phases here.

PR #634, opened 2026-03-12
Deduplicate CI build runs for internal PRs

hangleang:
This PR is to de-duplicate CI Build runs on internal PR so the `pull_request` event is skipped while the `push` event already handled it

**DISCLOSE**: AI-assisted coding

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,11 @@ name: CI Build
 
 on: [push, pull_request]
 
+# Cancel in-progress runs for the same ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/develop' }}
+
 env:
   TEST_CASES: |
     consensus spec tests mainnet fulu empty block transition
@@ -9,6 +14,8 @@ env:
 
 jobs:
   build:
+    # Skip pull_request events for same-repo branches (push already covers them)
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,17 @@
 name: CI Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "**"
+  pull_request:
+    branches:
+      - "**"
```
Tumas: what are the effects of this specific change?

.github/workflows/ci.yml
```diff
@@ -182,7 +197,7 @@ jobs:
       - name: Setup nim for constantine backend
         uses: jiro4989/setup-nim-action@v2
         with:
-          nim-version: '2.0.2'
+          nim-version: "2.0.2"
```
Tumas, line 194: is this change needed?

.github/workflows/ci.yml
```diff
@@ -1,6 +1,17 @@
...
+  push:
+    branches:
+      - "**"
+  pull_request:
+    branches:
+      - "**"
+
+# Cancel in-progress runs for the same ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
ArtiomTr, line 7: `github.ref` would default to `refs/heads/develop` in develop branch, so newer commits on develop branch will abort CI runs from previous commits. This is probably not what you wanted?

.github/workflows/ci.yml
hangleang: Claude said _"You're absolutely right!"_ :) It means the same thing

.github/workflows/ci.yml
hangleang: It just lint

.github/workflows/ci.yml
hangleang: you're right! CI runs on develop should not be cancelled. I've changed it in https://github.com/grandinetech/grandine/pull/634/changes/6b59e7603ab76f9389e1459147ecb1ef918358e0

ArtiomTr (approved):

PR #654, opened 2026-04-02
fix: use deposit_requests_start_index in Gloas process_operations

DELENE-TCHIO-ROMUALD:
The Gloas deposit count validation at `gloas/block_processing.rs:705-715`
  uses the pre-Electra formula (`deposit_count - eth1_deposit_index`) which
  ignores `deposit_requests_start_index`.

  The Electra implementation (`electra/block_processing.rs:499-524`) correctly
  uses `deposit_requests_start_index` to disable the former deposit mechanism
  once all prior deposits are processed, per the consensus spec
  (gloas/beacon-chain.md line 1157-1158). The Gloas version should use the
  same formula.

  Without this fix, Grandine would compute a wrong expected deposit count
  for Gloas blocks, causing a consensus split.

  AI Assistance Disclosure: This bug was identified with the assistance
  of Claude Code during a cross-client security audit. The fix was
  reviewed and tested manually .

### DIFF
--- a/transition_functions/src/gloas/block_processing.rs
+++ b/transition_functions/src/gloas/block_processing.rs
@@ -689,6 +689,7 @@ pub fn process_execution_payload_bid<P: Preset>(
     Ok(())
 }
 
+#[expect(clippy::too_many_lines)]
 pub fn process_operations<P: Preset, V: Verifier, B>(
     config: &Config,
     pubkey_cache: &PubkeyCache,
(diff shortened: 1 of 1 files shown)


hangleang (approved):
Thanks for catching this. This was a mistake in the initial Gloas implementation — I accidentally ported the `process_operation` from pre-Electra. The fix aligns with the spec at [gloas/beacon-chain.md:1157-1158](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.4/specs/gloas/beacon-chain.md?plain=1#L1198). LGTM

transition_functions/src/gloas/block_processing.rs
```diff
@@ -689,6 +689,7 @@ pub fn process_execution_payload_bid<P: Preset>(
     Ok(())
 }
 
+#[allow(clippy::too_many_lines)]
```
hangleang:
```suggestion
#[expect(clippy::too_many_lines)]
```

hangleang (changes requested):
just a nit with keyword

transition_functions/src/gloas/block_processing.rs
hangleang: you can verify by run this script https://github.com/grandinetech/grandine/blob/566c423d45af59ada3cd11b0c476f2b221d2d40d/.github/workflows/ci.yml#L87

hangleang (approved):

PR #656, opened 2026-04-03
Add Fast Confirmation Rule (FCR) implementation

bomanaps:
Implements the Fast Confirmation Rule (FCR) for the fork choice store, including the confirmation logic, chain safety checks, FFG data computation, and HTTP API integration for the safe block ID. Adds FCR-specific tests covering confirmation advancement, staleness reversion, and the disabled fallback path.

https://github.com/ethereum/consensus-specs/pull/4747


features/src/lib.rs
```diff
@@ -43,6 +43,7 @@ pub enum Feature {
     DumpBeaconBlocks,
     DumpBeaconStates,
     DumpBlobSidecars,
+    FastConfirmationRule,
```
ArtiomTr: probably not a good idea to have FCR as a feature: cli flag would look more logical there

fork_choice_control/src/extra_tests.rs
```diff
@@ -2522,3 +2523,270 @@ fn reorganizing_due_to_invalidation_sends_notifications_if_common_ancestor_is_un
...
+
+/// RAII guard that serializes FCR tests and disables `FastConfirmationRule` on drop.
+///
+/// `Feature::FastConfirmationRule` is a global flag. Without serialization, two FCR
+/// tests running concurrently can race: one test's `drop` disables the flag while the
+/// other test is still running, causing `confirmed_root()` to fall back to
+/// `justified_checkpoint.root` and break assertions.
+///
+/// Holding `FCR_LOCK` ensures at most one FCR test runs at a time.
+static FCR_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
ArtiomTr: so that's one of the reasons why FCR shouldn't be implemented as a feature -- if FCR would be a flag, you wouldn't need any of the locking stuff in the first place

fork_choice_store/src/fast_confirmation.rs
```diff
@@ -0,0 +1,427 @@
+/// Fast Confirmation Rule (FCR) constants, types, and free functions.
+///
+/// See the [FCR specification](https://github.com/ethereum/consensus-specs/blob/main/specs/phase0/fast-confirmation.md).
```
ArtiomTr: this link is dead?

bomanaps: https://github.com/ethereum/consensus-specs/pull/4747. sorry i will update and address this

fork_choice_store/src/fast_confirmation.rs
```diff
@@ -0,0 +1,433 @@
...
+use types::{
+    phase0::{
+        containers::Checkpoint,
+        primitives::{Epoch, Gwei, H256, Slot},
+    },
+    preset::Preset,
+};
+
+/// Assumed maximum percentage of Byzantine validators among the validator set.
+pub const CONFIRMATION_BYZANTINE_THRESHOLD: u64 = 25;
```
ArtiomTr: looks like this thing should be in config, as per spec?

https://github.com/mkalinin/eth2.0-specs/blob/5a58813bc737b02c001bc38ae41d75bb9131f511/configs/mainnet.yaml#L228

fork_choice_store/src/lib.rs
```diff
@@ -95,6 +95,7 @@ pub use crate::{
 mod blob_cache;
 mod data_column_cache;
 mod error;
+pub mod fast_confirmation;
```
ArtiomTr: don't think this should be `pub mod` - probably `mod`, and re-export needed members via `pub use` at the top of file

runtime/src/grandine_args.rs
```diff
@@ -125,6 +125,10 @@ pub struct GrandineArgs {
     #[clap(long, value_delimiter = ',')]
     features: Vec<Feature>,
 
+    /// Enable the Fast Confirmation Rule for single-slot block confirmation
+    #[clap(long)]
+    fast_confirmation_rule: bool,
```
ArtiomTr: this probably should go into either chain_options, or beacon_node_options

fork_choice_store/src/store.rs
```diff
@@ -4498,4 +4592,638 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
             self.current_slot_attestations.len(),
         );
     }
+
+    // =========================================================================
+    // Fast Confirmation Rule (FCR) — spec: fast-confirmation.md
+    // =========================================================================
```
ArtiomTr: please remove this comment

fork_choice_store/src/store.rs
```diff
@@ -4498,4 +4592,638 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+
+    /// spec: `get_equivocation_score(store, state, from_slot, to_slot)`
+    ///
+    /// Sums the active balances of validators that both appear in a committee
+    /// between `from_slot..=to_slot` AND are in `self.equivocating_indices`.
+    ///
+    /// Callers are responsible for the short-circuit when `equivocating_indices` is empty.
```
ArtiomTr: Please format comments as others in this file, i.e.:

https://github.com/grandinetech/grandine/blob/4607b7aba850327231c4027d5e88909a5275aec3/fork_choice_store/src/store.rs#L1863-L1867

fork_choice_store/src/store.rs
```diff
@@ -257,6 +258,25 @@ pub struct Store<P: Preset, S: Storage<P>> {
...
+    fcr_prev_gu_checkpoint: Checkpoint,
+    // spec: `previous_slot_head`
+    fcr_prev_slot_head: H256,
+    // spec: `current_slot_head`
+    fcr_curr_slot_head: H256,
+    // Cached result of compute_honest_ffg_support_for_current_target, refreshed each slot.
+    // Avoids redundant validator-set scans when the FFG gate is queried multiple times.
+    fcr_honest_ffg_support: Gwei,
+    // Cached total active balance from the current balance source, refreshed each slot.
+    fcr_ffg_total_active_balance: Gwei,
```
ArtiomTr: I think this is not the place for those fields. AFAIK, the FCR shouldn't affect forkchoice logic. For this purpose, the spec defines separate FCR store, that has only *readonly* access to fork-choice store. https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/fast-confirmation.md#fastconfirmationstore

The best solution would probably be to split FCR into its own store

fork_choice_store/src/store.rs
```diff
@@ -1030,11 +1059,56 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
     /// [`get_safe_execution_payload_hash`](https://github.com/ethereum/consensus-specs/blob/v1.3.0/fork_choice/safe-block.md#get_safe_execution_payload_hash)
```
ArtiomTr: this link needs to be updated - currently it points to an outdated spec version

fork_choice_store/src/store.rs
```diff
@@ -1030,11 +1059,56 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
     #[must_use]
     pub fn safe_execution_payload_hash(&self) -> ExecutionBlockHash {
+        if self.store_config.fast_confirmation_rule && self.fcr_confirmed_root != H256::zero() {
+            if let Some(hash) = self
+                .chain_link(self.fcr_confirmed_root)
+                .and_then(ChainLink::execution_block_hash)
+            {
+                return hash;
+            }
```
ArtiomTr: that doesn't look right - it shouldn't fallback to justified block root if confirmed root not found. You should probably jsut do the `.unwrap_or_default()`, instead of pattern matching

fork_choice_store/src/store.rs
```diff
@@ -1030,11 +1059,56 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
             .and_then(ChainLink::execution_block_hash)
             .unwrap_or_default()
     }
 
+    /// Returns the current confirmed block root.
+    ///
+    /// When `store_config.fast_confirmation_rule` is enabled, this is the FCR-confirmed root.
+    /// Otherwise falls back to the justified checkpoint root.
+    #[must_use]
+    pub fn confirmed_root(&self) -> H256 {
```
ArtiomTr: why confirmed root falls back to justified checkpoint? I think there is no such thing as "confirmed root", outside of FCR context?

ArtiomTr: also noticed that sometimes the finalized root is treated as confirmed, so probably justified root is too weak assumption there

fork_choice_store/src/store.rs
```diff
@@ -1030,11 +1059,56 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+    #[must_use]
+    pub fn fcr_prev_slot_head(&self) -> H256 {
+        self.fcr_prev_slot_head
+    }
+
+    /// Returns `store.current_slot_head`.
+    #[must_use]
+    pub fn fcr_curr_slot_head(&self) -> H256 {
+        self.fcr_curr_slot_head
+    }
```
ArtiomTr: don't think all of those should be public?

ArtiomTr (commented):
Haven't reviewed everything yet, but some places need major refactoring. Specifically, the biggest issue is that FCR lives in forkchoice store, but probably would be better living on its own.

Also, I don't see spec tests updated - those are required to be wired & passing, before merging

UPD: looks like formatting check was failing - all lints also need to be passing, before merging. For convenience, you can run them locally, with `./scripts/ci/clippy.bash --deny warnings`

PR #657, opened 2026-04-06
Refactor `validate_data_column_sidecar_with_state` function

hangleang:
This PR merge Fulu and Gloas data column sidecar validation into a single unified method to avoid converting between `Arc<DataColumnSidecar>` and its underlying type

### DIFF
--- a/fork_choice_store/src/store.rs
+++ b/fork_choice_store/src/store.rs
@@ -54,16 +54,10 @@ use types::{
         primitives::{BlobIndex, KzgCommitment},
     },
     electra::containers::IndexedAttestation as ElectraIndexedAttestation,
-    fulu::{
-        containers::{DataColumnIdentifier, DataColumnSidecar as FuluDataColumnSidecar},
-        primitives::ColumnIndex,
-    },
+    fulu::{containers::DataColumnIdentifier, primitives::ColumnIndex},
     gloas::{
         consts::BUILDER_INDEX_SELF_BUILD,
-        containers::{
-            DataColumnSidecar as GloasDataColumnSidecar, ProposerPreferences,
-            SignedExecutionPayloadBid,
-        },
+        containers::{ProposerPreferences, SignedExecutionPayloadBid},
         primitives::BuilderIndex,
     },
     nonstandard::{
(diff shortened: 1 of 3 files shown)


fork_choice_store/src/store.rs
```diff
@@ -2344,263 +2368,147 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
 
-        // [IGNORE] A valid block for the sidecar's `slot` has been seen (via gossip or non-gossip sources).
-        // A client MUST queue the sidecar for processing once the block is received or retrieved.
-        let Some(block) = self.block(block_root).map(WithStatus::value) else {
-            return Ok(DataColumnSidecarAction::DelayUntilState(
-                data_column_sidecar,
-                block_root,
-            ));
-        };
+            if !origin.is_from_back_sync() {
```
povi: Duplicate check `!origin.is_from_back_sync()`. There is a code block with an identical check just above this.

hangleang: updated in https://github.com/grandinetech/grandine/commit/9fc14ae4279c92afb37b103ad2a2c93a0315a23d

fork_choice_store/src/store.rs
```diff
@@ -2344,263 +2368,145 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
-
-        // [REJECT] The sidecar is valid as verified by verify_data_column_sidecar(sidecar)
-        ensure!(
-            verify_data_column_sidecar(&self.chain_config, &data_column_sidecar, kzg_commitments),
-            Error::DataColumnSidecarInvalid {
-                data_column_sidecar
-            },
-        );
-
-        // [REJECT] The sidecar is for the correct subnet -- i.e. compute_subnet_for_data_column_sidecar(sidecar.index) == subnet_id.
```
povi, line 2579: Is this validation is not needed anymore?

hangleang: it is already checked at https://github.com/hangleang/grandine/blob/9fc14ae4279c92afb37b103ad2a2c93a0315a23d/fork_choice_store/src/store.rs#L2357-L2369

PR #662, opened 2026-04-08
Validate payload envelope through processing

hangleang:


fork_choice_store/src/store.rs
```diff
@@ -3063,6 +3125,46 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
             ));
         }
 
+        // > Add new state for this payload to the store
+        Ok(ExecutionPayloadEnvelopeAction::Accept(envelope, state))
+    }
+
+    pub fn validate_execution_payload_envelope_with_state(
```
povi, line 3132: This method is very similar to `validate_execution_payload_envelope_with_custom_state_transition` and is used only in back sync. Maybe we could reuse this method for back sync too? Or call `validate_execution_payload_envelope_with_custom_state_transition` method inside this one?

hangleang: We could fake state transition to just return the state

povi (approved):

PR #667, opened 2026-04-16
Feature/consensus specs v1.7.0 alpha4

Tumas:
Depends on https://github.com/grandinetech/grandine/pull/663 and https://github.com/grandinetech/grandine/pull/666 being merged first


fork_choice_store/src/store.rs
```diff
@@ -3445,6 +3525,45 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
         self.apply_balance_differences(differences)
     }
 
+    pub fn observe_gossip_attester(&mut self, validator_index: ValidatorIndex, epoch: Epoch) {
+        if let Ok(index) = usize::try_from(validator_index) {
+            let bitfield = self.seen_gossip_attesters.entry(epoch).or_default();
+
+            if bitfield.len() <= index {
+                bitfield.resize(index + 1, false);
```
povi, line 3538: I don't like this. If mainnet has >1M validators, does that mean it can do a million vector resizes per epoch? Please [reserve](https://docs.rs/bitvec/1.0.1/bitvec/vec/struct.BitVec.html#method.reserve) bitvec so it is as big as total validators count on initialization.

povi: But keep this, since new validators can be added too

povi (approved):

PR #671, opened 2026-04-22
Add `#![no_main]` to suppress duplicate entry point error

hangleang:
`risc0_zkvm::guest::entry!(main)` defines its own entry point, so a recent risc0 version update stopped auto-suppressing the standard one, causing a duplicate main function error at compile time


ArtiomTr (approved):
LGTM 👍

PR #673, opened 2026-04-23
Delay attestations with payload_present=1 until envelope received

Subhasish-Behera:
Implements the Gloas gossip rule: attestations with `data.index == 1` voting for a post-Gloas block are delayed until the local envelope for that block is received.      


                                                                                                                                                                                                                                                                                                
Spec Reference:   
https://github.com/ethereum/consensus-specs/blob/a6687a2de15a0cbd92b052ce3a7727f83ef49b53/specs/gloas/p2p-interface.md?plain=1#L244-L248

                                                                        
  https://github.com/ethereum/consensus-specs/blob/a6687a2de15a0cbd92b052ce3a7727f83ef49b53/specs/gloas/p2p-interface.md?plain=1#L529-L533
  
Note: not affected by deferred payload state apply refactor. as this is a by-root request(not by range) and part of the fork choice calculation


fork_choice_control/src/mutator.rs
```diff
@@ -1942,12 +1981,14 @@ where
         }
     }
 
-    #[expect(clippy::too_many_lines)]
+    #[expect(clippy::too_many_arguments, clippy::too_many_lines)]
```
Subhasish-Behera, line 1984: should i use a separate identifier out of 
`  beacon_block_root: H256,
  builder_index: BuilderIndex` . 
currently bypassing clippy with `clippy::too_many_arguments`

PR #677, opened 2026-05-05
Decouple validator from beacon node via traits

bomanaps:
Refactored the validator so it talks to the beacon node through a small set of trait objects (BeaconChainReader, BeaconDutyEndpoints, BeaconPublisher, BeaconEventStream) instead of poking  pools, channels, and the controller directly. Behavior is identical every call still resolves to the same in-process function via the new LocalBeaconClient impl but the seam is now in place to drop in a remote HTTP client later without touching the validator again.


validator/src/validator.rs
```diff
@@ -141,6 +144,10 @@ pub struct Validator<P: Preset, W: Wait> {
     validator_config: Arc<ValidatorConfig>,
     block_producer: Arc<BlockProducer<P, W>>,
     controller: ApiController<P, W>,
+    reader: Arc<dyn BeaconChainReader<P>>,
+    duties: Arc<dyn BeaconDutyEndpoints<P>>,
+    publisher: Arc<dyn BeaconPublisher<P>>,
+    events: Arc<dyn BeaconEventStream<P>>,
```
ArtiomTr: why do we need 4 different objects here?

beacon_node_client/src/local.rs
```diff
@@ -0,0 +1,896 @@
...
+        Ok(AttestationData {
+            slot,
+            index: committee_index,
+            beacon_block_root: block_root,
+            source: state.current_justified_checkpoint(),
+            target,
+        })
+    }
+
+    async fn produce_block_v3(
```
ArtiomTr: why is that `v3`? looks like we don't have `v1` or `v2`

beacon_node_client/src/traits.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    async fn is_registered_validator(&self, validator_index: ValidatorIndex) -> bool;
+
+    async fn registered_validator_indices(&self) -> HashSet<ValidatorIndex>;
+
+    async fn no_prepared_proposers(&self) -> bool;
+
+    async fn prepared_proposer_indices(&self) -> Vec<ValidatorIndex>;
+
+    fn kzg_backend(&self) -> KzgBackend;
```
ArtiomTr: why is that needed here?

PR #679, opened 2026-05-07
fix: include block hash in Gloas genesis block

hangleang:
implement https://github.com/ethpandaops/eth-beacon-genesis/pull/77


povi (approved):
LGTM

PR #691, opened 2026-05-14
Add missing SSE events

### LINKED ISSUE #800: [Glamsterdam] add missing SSE events
https://investigations.ethpandaops.io/2026-07/glamsterdam-event-client-matrix/

hangleang:
Repurpose the PR to add missing SSE events as of report in https://investigations.ethpandaops.io/2026-07/glamsterdam-event-client-matrix/

Closes #800


povi (approved):

fork_choice_control/src/mutator.rs
```diff
@@ -3629,6 +3637,16 @@ where
         );
     }
 
+    fn send_head_events(&self, head: &ChainLink<P>) {
+        self.event_channels
+            .send_head_event(head, |head| self.calculate_dependent_roots(head));
+
+        self.event_channels
+            .send_head_v2_event(head, self.store.head_payload_status(), |head| {
+                self.calculate_dependent_roots(head)
```
povi: You could just calculate dependent roots bundle beforehand and pass it as an argument to `send_head_v2_event`. Then there would be no need to pass anonymous functions as a parameter, and it would do one `calculate_dependent_roots` call instead of two.

Other than that, LGTM

hangleang: I will merge this PR after the build pass

PR #693, opened 2026-05-15
Add target_gas_limit to PayloadAttributes

hangleang:
Implement https://github.com/ethereum/consensus-specs/pull/5235

### DIFF
--- a/block_producer/src/block_producer.rs
+++ b/block_producer/src/block_producer.rs
@@ -49,9 +49,12 @@ use types::{
             SyncAggregate,
         },
     },
-    bellatrix::containers::{
-        BeaconBlock as BellatrixBeaconBlock, BeaconBlockBody as BellatrixBeaconBlockBody,
-        ExecutionPayload as BellatrixExecutionPayload,
+    bellatrix::{
+        containers::{
+            BeaconBlock as BellatrixBeaconBlock, BeaconBlockBody as BellatrixBeaconBlockBody,
+            ExecutionPayload as BellatrixExecutionPayload,
+        },
+        primitives::Gas,
     },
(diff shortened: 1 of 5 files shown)


block_producer/src/block_producer.rs
```diff
@@ -2253,6 +2258,14 @@ impl<P: Preset, W: Wait> BlockBuildContext<P, W> {
             })
     }
 
+    async fn gas_limit(&self) -> Result<Gas> {
```
hangleang, line 2261: I guess we should introduce new version of `POST /eth/v1/validator/prepare_beacon_proposer` to allow validator client submit their prefered gas limit to beacon node as well

cc @povi @Tumas

povi: There is this endpoint, which we already implement (also whole API): https://ethereum.github.io/keymanager-APIs/?urls.primaryName=dev#/Gas%20Limit/setGasLimit

hangleang: yes, I see. but how can we load it in block building like [`fee_recipient`](https://github.com/grandinetech/grandine/blob/0db0d663b8f870ab743ff2fe76135439470156d8/block_producer/src/block_producer.rs#L1952)?

povi: We already pass preferred gas limit to external builders: https://github.com/grandinetech/grandine/blob/develop/validator/src/validator.rs#L2295

You can do something like that in block producer.

hangleang: Okay, that's what I did. So it's fine

Tumas (approved):

PR #698, opened 2026-05-15
Deprecate unsafe arithmetics

Tumas:


types/src/config.rs
```diff
@@ -1074,13 +1074,16 @@ impl Config {
     }
 
     #[must_use]
-    pub const fn custody_size<P: Preset>(&self, sampling_column_count: u64) -> u64 {
-        sampling_column_count.saturating_div(self.columns_per_group::<P>())
+    pub const fn custody_size<P: Preset>(&self, sampling_column_count: u64) -> Option<u64> {
+        sampling_column_count.checked_div(self.columns_per_group::<P>())
```
povi: I think this and other config values should have an `expect` internally. E. g. this one could be `.expect("columns_per_group should never be zero")`

types/src/config.rs
```diff
@@ -1074,13 +1074,16 @@ impl Config {
...
     }
 
     #[must_use]
     pub const fn columns_per_group<P: Preset>(&self) -> u64 {
-        P::NumberOfColumns::U64.saturating_div(self.number_of_custody_groups)
+        match P::NumberOfColumns::U64.checked_div(self.number_of_custody_groups) {
+            Some(v) => v,
+            None => 0,
```
povi: It should never be zero, as `NUMBER_OF_COLUMNS` or `NUMBER_OF_CUSTODY_GROUPS` should never be a zero either. Use `expect` here instead of `match`

povi (approved):

PR #707, opened 2026-05-25
Fix nethermind integration memory leak

ArtiomTr:
Fixed two memory leak issues in C bindings:
* implemented proper cleanup for method arguments - they are no longer moved through FFI boundary, but passed by reference;
* implemented proper cleanup for CVec<T> items.


povi (approved):
LGTM

PR #718, opened 2026-06-01
Add payload present timely store

### LINKED ISSUE #710: [Glamsterdam] Dual deadlines
Implement dual deadlines.

hangleang:
Implement https://github.com/ethereum/consensus-specs/pull/5212 and rename PTC payload timely votes predicate function to avoid confusion with the local view. 

Close #710

### DIFF
--- a/fork_choice_store/src/store.rs
+++ b/fork_choice_store/src/store.rs
@@ -189,6 +189,7 @@ pub struct Store<P: Preset, S: Storage<P>> {
     // - Obtaining active balances from the justified state requires it to be in the right epoch.
     checkpoint_states: HashMap<Checkpoint, Arc<BeaconState<P>>>,
     payloads: HashSet<H256>,
+    timely_payloads: HashSet<H256>,
     // This field is used to track which PTC members have cast a valid vote for a given block root.
     payload_vote: HashMap<H256, BitVector<P::PtcSize>>,
     payload_timeliness_vote: HashMap<H256, BitVector<P::PtcSize>>,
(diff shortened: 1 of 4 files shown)


fork_choice_store/src/store.rs
```diff
@@ -5215,6 +5223,19 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
     }
 
+    // Return whether the current tick is before the deadline defined by `due_bps`
+    // basis points into the slot.
+    fn is_before_due_bps_deadline(&self, due_bps: u64) -> bool {
+        let ticks_per_slot = u64::try_from(TickKind::ticks_per_slot::<P>(
+            &self.chain_config,
+            self.slot(),
+        ))
+        .unwrap_or(u64::MAX);
```
povi: I am not a fan of these silent failures, despite how unlikely they are. Please either return `Result<bool>` from a method, or add an `expect` with an explicit reason you are sure why this won't error ever.

fork_choice_store/src/store.rs
```diff
@@ -5215,6 +5223,19 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+        let tick_index = self.tick.kind as u64;
+
+        (tick_index + 1) * BASIS_POINTS <= due_bps * ticks_per_slot
```
povi: We are now not allowing unchecked arithmetics in the app. Please check commit https://github.com/grandinetech/grandine/commit/0de31c0ffe36a36e4a003ab69cbec4981f78e53a where this is changed

fork_choice_store/src/store.rs
hangleang: updated

povi (approved):

PR #719, opened 2026-06-01
Reintroduce networking spec tests

Tumas:
- Re-enable gossip spec tests
- Add `AnchorBlock` and `AnchorState` enums for test-specific behaviour
- Fix combined Attestation SSZ deserialization


fork_choice_store/src/misc.rs
```diff
@@ -279,13 +356,14 @@ pub enum BlockOrigin {
     Own,
     Persisted,
     Api(Option<Sender<Result<ValidationOutcome>>>),
+    Test(GossipId, BlsSetting, StateRootPolicy),
```
povi, line 341: Is there a reason why cannot this be under `#[cfg(test)]`? Then there would be no need to load `BlsSetting` to prod code and move `spec_test_utils` from dev-dependencies to dependencies

Tumas: Networking tests need a context initialized with pools, the controller, and validator services to mirror the real application, so they live in the validator crate. Running them compiles validator with cfg(test), but workspace dependencies like fork_choice_store (where this message is defined) are compiled without it, so #[cfg(test)] items here are inactive during those tests — see rust-lang/rust#59168. This could likely be worked around with a feature flag, but unclear if it's worth it.

PR #726, opened 2026-06-01
EIP-8045: Exclude slashed validators from proposing

povi:
Implements https://github.com/grandinetech/grandine/issues/725


hangleang (approved):
looks good

PR #748, opened 2026-06-10
Introduce `is_previous_slot_payload_decision` predicate

povi:


hangleang (approved):
Looks good, but I haven't fully verified the change yet

PR #765, opened 2026-06-19
Update consensus-spec-tests for `glamsterdam-devnet-6`

### LINKED ISSUE #764: [Glamsterdam] Update consensus-spec-tests to v1.7.0-alpha.11
Tracking consensus spec changes in https://notes.ethereum.org/@ethpandaops/glamsterdam-devnet-6

hangleang:
Implement #764

### DIFF
--- a/eth1_api/src/eth1_api/http_api.rs
+++ b/eth1_api/src/eth1_api/http_api.rs
@@ -335,7 +335,7 @@ impl Eth1Api {
             }
             (
                 ExecutionPayload::Gloas(payload),
-                Some(ExecutionPayloadParams::Electra {
+                Some(ExecutionPayloadParams::Gloas {
                     versioned_hashes,
                     parent_beacon_block_root,
                     execution_requests,
@@ -1046,7 +1046,7 @@ mod tests {
         assert_eq!(payload.value.phase(), Phase::Deneb);
         assert_eq!(
             payload.execution_requests,
-            Some(ExecutionRequests::default())
+            Some(ExecutionRequests::default().into())
         );
 
         Ok(())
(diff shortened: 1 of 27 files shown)


types/src/combined.rs
```diff
@@ -968,7 +970,7 @@ impl<P: Preset> BeaconBlock<P> {
         mut self,
         execution_requests: Option<ExecutionRequests<P>>,
     ) -> Self {
-        let Some(execution_requests) = execution_requests else {
+        let Some(ExecutionRequests::Electra(execution_requests)) = execution_requests else {
```
povi, line 973: Shouldn't it also handle Gloas execution requests?

eth1_api/src/eth1_api/http_api.rs
```diff
@@ -335,7 +335,7 @@ impl Eth1Api {
             }
             (
                 ExecutionPayload::Gloas(payload),
-                Some(ExecutionPayloadParams::Electra {
+                Some(ExecutionPayloadParams::Gloas {
```
povi, line 338: I don't see any similar updates to `embed_api`. Did you check it? If there are places in it that needs updating, but you're unsure how to do it, please add a `TODO(Gloas)`

povi (approved):

types/src/combined.rs
hangleang: We already moved `execution_requests` to payload envelope out of beacon block body, and we don't have a code path to call the function with Gloas block either

eth1_api/src/eth1_api/http_api.rs
hangleang: Created a PR here https://github.com/grandinetech/grandine/pull/784

PR #804, opened 2026-07-10
Wait for proposer preferences broadcasting to finish

hangleang:
The `minimal_rapid_upgrade .. all_keys .. built_in_validator` snapshot test failed on macOS CI:
```
json atom at path ".validator_to_p2p[1].PublishProposerPreferences" is missing from lhs
json atom at path ".validator_to_p2p[2..16]" is missing from lhs
```
  The run produced only `UpdateDataColumnSubnets` + `PublishBeaconBlock` and
  dropped all 15 `PublishProposerPreferences` messages the snapshot expects.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -2751,115 +2751,99 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
     /// Broadcasts proposer preferences for the next epoch's slots where we are proposing.
     #[instrument(level = "debug", skip_all)]
     async fn broadcast_proposer_preferences(&self, slot_head: &SlotHead<P>) {
-        let chain_config = self.chain_config.clone_arc();
-        let proposer_configs = self.proposer_configs.clone_arc();
-        let signer = self.signer.clone_arc();
-        let p2p_tx = self.p2p_tx.clone();
-        let beacon_state = slot_head.beacon_state.clone_arc();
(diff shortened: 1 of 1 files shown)


validator/src/validator.rs
```diff
@@ -2731,114 +2731,107 @@ impl<P: Preset, W: Wait + Sync> Validator<P, W> {
     /// Broadcasts proposer preferences for the next epoch's slots where we are proposing.
     #[instrument(level = "debug", skip_all)]
     async fn broadcast_proposer_preferences(&self, slot_head: &SlotHead<P>) {
```
povi, line 2753: Have you tried to add `wait_group: W` param to `broadcast_proposer_preferences` and then `drop(wait_group)` on the last line inside `tokio::spawn`? This way it ensures that task is finished before checking test results (`wait_for_tasks` method call waits until all references to `wait_group` are dropped before continuing).

hangleang: let me try

hangleang: added wait_group in f49c32ba24cdf85dd5916bf0dabe44c3ac193dec

hangleang: CI build failed again, it still produced non-deterministic behavior

povi: Okay. Then previous approach may be better

PR #810, opened 2026-07-14
Feature/stable containers

povi:


types/src/combined.rs
```diff
@@ -2138,6 +2143,14 @@ impl<P: Preset> SszSize for Attestation<P> {
     ]);
 }
 
+// TODO(gloas): make `Attestation::Gloas` a first-class variant instead of
```
hangleang, line 2146: There is the missing piece for interop

povi: I'm going to add that with a separate PR

PR #813, opened 2026-07-15
Gloas attestations

povi:
Implements #812

### DIFF
--- a/fork_choice_store/src/store.rs
+++ b/fork_choice_store/src/store.rs
@@ -24,7 +24,7 @@ use hash_hasher::HashedMap;
 use helper_functions::{
     accessors, electra,
     error::SignatureKind,
-    misc, phase0, predicates,
+    gloas, misc, phase0, predicates,
     signing::SignForSingleFork as _,
     slot_report::NullSlotReport,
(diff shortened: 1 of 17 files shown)


fork_choice_store/src/store.rs
```diff
@@ -2827,6 +2827,24 @@ impl<P: Preset, S: Storage<P>> Store<P, S> {
...
+                    predicates::validate_constructed_indexed_attestation(
+                        &self.chain_config,
+                        &self.pubkey_cache,
+                        target_state,
+                        &indexed_attestation,
+                        SingleVerifier,
+                    )?;
+                }
+
+                Ok(AttestingIndices::Electra(
```
hangleang: This will convert into `ContiguousList`, not stable container

we need `AttestingIndices::Gloas` as `ProgressiveList[ValidatorIndex]`

povi: Good suggestion, I went a step further ahead and removed `enum IndexedAttestation`, as `ProgressiveList` is enough (it's not bounded)

hangleang (approved):

Issue #769, opened 2026-06-24
[Glamsterdam] Staked Builder API


hangleang:
Implement client side of https://github.com/ethereum/builder-specs/pull/138 into block production

NikhilSharmaWe:
Can I take this up? If no one has been assigned yet.

NikhilSharmaWe:
@hangleang I'd like to take this issue if it's still unassigned.

After reviewing `glamsterdam-devnet-7`, most of the Gloas consensus work already looks in place: execution payload bid validation, envelope handling, fork choice integration, and proposer preferences gossip (publish/receive on the validator/p2p side).

What still seems missing is the Gloas builder API HTTP client in `builder_api`:
- `getExecutionPayloadBid`
- `submitBuilderPreferences`
- `submitSignedBeaconBlock`

…and wiring that into block production for the HTTP staked-builder path. The branch already supports gossip-based bid selection and self-build with `SignedExecutionPayloadBid`, so this looks more like adding the builder-specs#138 client flow than replacing the pre-Gloas blinded-block path.

A couple of questions before I start:
1. On devnet-7, should external builders use this HTTP path, or is gossip bid reception enough for now?
2. When using the HTTP path, should the proposer stop gossiping the block after `submitSignedBeaconBlock` returns 202 (per spec)?

Happy to land this incrementally, starting with a `builder_api`-only PR. Please let me know if that approach works.

hangleang:
@NikhilSharmaWe please complete this PR first #801

NikhilSharmaWe:
@hangleang I've addressed the failing CI in #801.

Could you please share your thoughts on the questions above when you have a chance? I'd like to confirm my understanding of the expected scope before I start working on this.

hangleang:
> 1. On devnet-7, should external builders use this HTTP path, or is gossip bid reception enough for now?

Not a hard requirement for devnet, it's nice to have it test

> 2. When using the HTTP path, should the proposer stop gossiping the block after submitSignedBeaconBlock returns 202 (per spec)?

the endpoint is just sent signed beacon block to the builder over http, proposer still need to gossip the block to the network for other validators to validate

PR #819, opened 2026-07-17
builder_api: add Gloas Staked Builder API client

NikhilSharmaWe:
Part of #769. Client side of the Gloas Staked Builder API ([ethereum/builder-specs#138](https://github.com/ethereum/builder-specs/pull/138)), aligned with the clarifications in [ethereum/builder-specs#165](https://github.com/ethereum/builder-specs/pull/165) (ACDT CL guidance: implement before merge).

## Changes

- Add `DOMAIN_REQUEST_AUTH` (`0x0B000001`) and the Gloas builder containers (`RequestAuthV1`, `SignedRequestAuthV1`, `BuilderPreferencesV1`, `BuilderPreferencesRequestV1`) to `builder_api`
- Add three HTTP client methods:
  - `get_execution_payload_bid` — POST `/eth/v1/builder/execution_payload_bid/{slot}/{parent_hash}/{parent_root}/{proposer_pubkey}` with a required `SignedRequestAuthV1` body (JSON or SSZ), required `Date-Milliseconds` and `X-Timeout-Ms` headers (timeout matches the reqwest timeout), and always `Eth-Consensus-Version`. 200 returns a bid, 204 means no bid. Gloas+ only; rejects earlier phases and `auth.message.slot` ≠ path `slot` before the request.
  - `submit_builder_preferences` — POST `/eth/v1/builder/builder_preferences/{validator_pubkey}` with `BuilderPreferencesRequestV1` (SSZ field order: `auth` then `preferences`), expects 202
  - `submit_signed_beacon_block` — POST `/eth/v1/builder/beacon_blocks`, expects 202
- Signing support for `RequestAuthV1` (`SignForAllForks`, genesis fork + zero genesis validators root, per spec `compute_domain(DOMAIN_REQUEST_AUTH)`)
- Prometheus histograms for the three calls

## Notes

- Client only. Wiring into block production is a follow-up (#822).
- Builder bid signature verification is deferred to the caller (needs beacon state / `state.builders()` by `builder_index`).
- After a 202 from `submitSignedBeaconBlock` the proposer must still gossip the beacon block; the builder publishes the payload envelope.
- Spec type names in ethereum/builder-specs#165 drop the `V1` suffix; Rust keeps `*V1` to match other `builder_api` containers. Wire format follows ethereum/builder-specs#165.

## Tests

- `cargo test -p builder_api --features blst` — httpmock coverage for all three endpoints (200/204/400/500, required auth body + timing / consensus-version headers, auth slot mismatch, non-2xx handling, pre-Gloas rejection), `DOMAIN_REQUEST_AUTH` signing-root check, and SSZ field-order lock for `BuilderPreferencesRequestV1`


builder_api/src/gloas/containers.rs
hangleang, line 1: just cosmetic, all new containers name has been removed the version suffix

builder_api/src/gloas/containers.rs
```diff
@@ -0,0 +1,131 @@
...
+    gloas::containers::SignedExecutionPayloadBid,
+    nonstandard::Phase,
+    phase0::primitives::{Gwei, Slot},
+    preset::Preset,
+};
+
+/// [`MAX_DATA_SIZE`] from `builder-specs` (Gloas).
+///
+/// [`MAX_DATA_SIZE`]: https://github.com/ethereum/builder-specs/blob/main/specs/gloas/builder.md#constants
+pub type MaxDataSize = U4096;
```
hangleang: it should be put in `consts.rs` i believe

builder_api/src/gloas/containers.rs
```diff
@@ -0,0 +1,131 @@
...
+
+#[derive(Debug, Deserialize)]
+#[serde(
+    bound = "",
+    deny_unknown_fields,
+    rename_all = "lowercase",
+    tag = "version",
+    content = "data"
+)]
+pub enum GetExecutionPayloadBidResponse<P: Preset> {
```
hangleang: this response type and its implementations should be put in `combined.rs`

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+                expected: StatusCode::OK,
+                received: response.status()
+            });
+        }
+
+        let bid_response = self
+            .parse_gloas_response::<GetExecutionPayloadBidResponse<P>>(response)
+            .await?;
+
+        let bid = bid_response.into_bid();
```
hangleang, line 598: returned bid isn't checked against request slot, parent_hash, parent_root as specified in https://github.com/ethereum/builder-specs/blob/main/specs/gloas/validator.md#validating-a-signedexecutionpayloadbid

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+        &self,
+        pubkey: PublicKeyBytes,
+        request_body: &BuilderPreferencesRequestV1,
+    ) -> Result<()> {
+        let _timer = self.metrics.as_ref().map(|metrics| {
+            metrics
+                .builder_submit_builder_preferences_times
+                .start_timer()
+        });
+
```
hangleang, line 640: why won't check slot like `get_execution_payload_bid` does?

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+        let url = self.url(&format!("/eth/v1/builder/builder_preferences/{pubkey:?}"))?;
+
+        let use_json = self.config.builder_api_format == BuilderApiFormat::Json;
+
+        debug_with_peers!("submitting builder preferences to {url}, use_json: {use_json}");
+
+        let request = self
+            .client
+            .post(url.into_url())
+            .header(ETH_CONSENSUS_VERSION, Phase::Gloas.as_ref());
```
hangleang: should derive phase from proposal slot

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+
+        let request = if use_json {
+            request.json(request_body)
+        } else {
+            request
+                .header(CONTENT_TYPE, APPLICATION_OCTET_STREAM.as_ref())
```
hangleang: do we need `Accept` header here as well?

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+        chain_config: &ChainConfig,
+        genesis_time: UnixSeconds,
+        block: &SignedBeaconBlock<P>,
+    ) -> Result<()> {
+        let _timer = self.metrics.as_ref().map(|metrics| {
+            metrics
+                .builder_submit_signed_beacon_block_times
+                .start_timer()
+        });
+
```
hangleang, line 697: why don't check block phase? maybe create a reusable function like `validate_phase` with corresponding error

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+            return Ok(());
+        }
+
+        bail!(BuilderApiError::UnexpectedStatusCode {
+            expected: StatusCode::ACCEPTED,
+            received: response.status()
+        })
+    }
+
+    async fn parse_gloas_response<T: DeserializeOwned + SszRead<Phase>>(
```
hangleang: this function almost identical to `parse_response` below, can we reuse the function?

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+             before next interval {next_interval:?}, use_json: {use_json}",
+        );
+
+        let block_root = block.message().hash_tree_root();
+        let slot = block.message().slot();
+
+        let request = self
+            .client
+            .post(url.into_url())
+            .timeout(remaining_time)
```
hangleang: this might be the wrong deadline for this endpoint. I guess this is copied from `post_blinded_block` pre-Gloas path which proposer require builder to response before attestation deadline, so it has the payload to construct and publish beacon block before deadline. but now builder is responsible to publish their payload, and this endpoint is just notify them to publish payload envelope by themselves through beacon node

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+
+        let date_ms = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .map(|timestamp| timestamp.as_millis())
+            .map_err(|error| anyhow::anyhow!("unable to calculate Date-Milliseconds: {error:?}"))?;
+
+        let request = self
+            .client
+            .post(url.into_url())
+            .timeout(REQUEST_TIMEOUT)
```
hangleang: the interval timeout should be put on this endpoint instead, as now proposer request bid from builder to choose on proposal hot path

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+            .header(DATE_MS_HEADER, format!("{date_ms}"))
+            .header(
+                X_TIMEOUT_MS_HEADER,
+                format!("{}", REQUEST_TIMEOUT.as_millis()),
```
hangleang: same here

builder_api/src/api.rs
```diff
@@ -462,6 +471,240 @@ impl Api {
...
+                auth_slot: signed_request_auth.message.slot,
+                path_slot: slot,
+            },
+        );
+
+        let url = self.url(&format!(
+            "/eth/v1/builder/execution_payload_bid/{slot}/{parent_hash:?}/{parent_root:?}/{pubkey:?}"
+        ))?;
+
+        let use_json = self.config.builder_api_format == BuilderApiFormat::Json;
```
hangleang: should have fallback to json format if builder doesn't support SSZ, same as other two endpoints

builder_api/src/gloas/containers.rs
NikhilSharmaWe: Done in 7293d08e

builder_api/src/gloas/containers.rs
NikhilSharmaWe: Done in 9dd626bd

builder_api/src/gloas/containers.rs
NikhilSharmaWe: Done in 298ce3bc

builder_api/src/api.rs
NikhilSharmaWe: Done in f98ab295. The returned bid’s slot, parent block hash and parent block root are now checked against the request

builder_api/src/api.rs
NikhilSharmaWe: Done in e5ba3898
auth.message.slot is treated as the proposal slot and rejected if it has already passed.

builder_api/src/api.rs
NikhilSharmaWe: Done in the same commit e5ba3898

builder_api/src/api.rs
NikhilSharmaWe: I left Accept unset intentionally because the success response is an empty 202; Content-Type still identifies the request-body encoding.

builder_api/src/api.rs
NikhilSharmaWe: Done in ba9600cc. `submit_signed_beacon_block` now rejects pre-Gloas blocks through `ensure_gloas_phase`.

builder_api/src/api.rs
NikhilSharmaWe: Done in 0e01a9ca. JSON/SSZ decoding was extracted into `decode_ssz_or_json_response` and reused. `parse_response` remains a wrapper for legacy builder responses because it also updates the separate `supports_block_ssz` cache.

builder_api/src/api.rs
NikhilSharmaWe: Fixed in a5867d4f. The signed-block notification now uses the fixed request timeout instead of the attestation-interval deadline.

builder_api/src/api.rs
NikhilSharmaWe: Done in 25b73d35
The bid request is bounded by the remaining time before the next interval, capped at one second.

builder_api/src/api.rs
NikhilSharmaWe: Done in same 25b73d35. X-Timeout-Ms and the reqwest timeout use the same computed millisecond value.

builder_api/src/api.rs
NikhilSharmaWe: Done for bids in e049bacf, then extended across all Gloas POST bodies in 0e01a9ca. A 415 retries once as JSON; 400 and other client errors are not retried.

builder_api/src/api.rs
hangleang: I don't think it is a good idea to do preliminary check the request slot against current slot tick

builder_api/src/api.rs
hangleang: can the function be more generic? like `validate_post_phase(phase, post_phase)` expect `phase` to be at least `post_phase`

Issue #795, opened 2026-07-09
Circuit-breaker for epbs


sauliusgrigaitis:
We need a custom circuit-breaker for (ebps). It's client specific implementation.

sauliusgrigaitis:
See ACDC #182 around 34:00.

PR #843, opened 2026-08-11
Add Gloas global circuit breaker

hangleang:
Implement #795 by reusing existing flags from pre-Gloas config for builder API, therefore no new flags are introduced.

This PR doesn't contains builder blacklisting mechanism, I will defer to the another PR


povi (approved):
LGTM

PR #850, opened 2026-08-17
Measure payload envelope timeliness at arrival

hangleang:
`handle_execution_payload_envelope` evaluated the payload deadline at the moment the
  mutator applied the envelope, everything between arrival and the evaluation is internal processing delay, we should not count it as delay on arrival.


povi (approved):
Please change target branch to `develop`

Issue #852, opened 2026-08-17
Add gas limit schedule


hangleang:
implement https://github.com/ethereum/consensus-specs/pull/5533

PR #854, opened 2026-08-18
Add gas limit schedule config

hangleang:
Implement #852

### DIFF
--- a/types/src/config.rs
+++ b/types/src/config.rs
@@ -15,7 +15,7 @@ use thiserror::Error;
 use typenum::Unsigned as _;
 
 use crate::{
-    bellatrix::primitives::Difficulty,
+    bellatrix::primitives::{Difficulty, Gas},
     fulu::containers::DataColumnsByRootIdentifier,
     nonstandard::{CustodyMode, Phase, Toption},
     phase0::{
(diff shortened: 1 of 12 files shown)


povi (approved):

types/src/config.rs
```diff
@@ -1144,6 +1158,25 @@ impl Config {
...
+    #[must_use]
+    pub fn get_scheduled_gas_limit(&self, epoch: Epoch) -> Option<Gas> {
+        if epoch < self.gloas_fork_epoch {
+            return None;
+        }
+
+        self.gas_limit_schedule
+            .iter()
+            .sorted_by(|a, b| b.epoch.cmp(&a.epoch))
+            .find_map(|entry| (epoch >= entry.epoch).then_some(entry.gas_limit))
```
povi, line 1177: Is this correct? This sorts entries from highest epoch to the lowest and then searches for first matching entry (and always gets the highest one)

hangleang: no, it sort descending, then find the first entry that at least match the current epoch, same as we do with blob schedule

povi: Ah, yes. My bad

PR #858, opened 2026-08-21
Updated SSZ spec runner

povi:


scripts/ci/consensus-spec-tests-coverage.rb
```diff
@@ -25,40 +25,70 @@
   tests/*/*/fast_confirmation/*/*/*/*.{ssz_snappy,yaml}
   tests/general/phase0/ssz_generic/compatible_unions/*/*/*.{ssz_snappy,yaml}
```
hangleang: it's now no longer in the consensus-spec-tests release, also the partial data column containers ssz tests

hangleang: also `Cargo.lock` isn't reflected

povi: Good catch!

ArtiomTr (approved):

PR #887, opened 2026-08-27
Replay persisted execution payload envelopes at restart

### LINKED ISSUE #886: Load payload envelopes from storage during restart
at restart, we should also replay persisted payload envelopes along with its beacon block in `process_unfinalized_blocks` otherwise, the node will need to range sync to advance to the head of the chain since post-Gloas block requires its parent payload to be available in order to import into fork choice.

hangleang:
Close #886 

This commit also add SSZ limits check on execution request operations and withdrawals length check in execution payload envelope validation. In p2p changes, fix stale peer custody info and request missing columns after forward synced at restart.

### DIFF
--- a/fork_choice_control/src/mutator.rs
+++ b/fork_choice_control/src/mutator.rs
@@ -514,14 +514,16 @@ where
             NullVerifier,
         );
 
-        // TODO: timeliness of a payload is not persisted, so we cannot know if it was seen before
-        // the deadline. We now assume it was, but we should persist that information as well.
+        let seen_before_deadline = self
+            .storage
+            .is_execution_payload_envelope_timely(block_root)?;
+
         self.handle_execution_payload_envelope(
             wait_group.clone(),
             result,
             origin,
             payload_envelope_identifier,
-            true,
+            seen_before_deadline,
             processing_timings,
             tracing::debug_span!("handle_persisted_envelope"),
         );
--- a/fork_choice_control/src/mutator.rs
+++ b/fork_choice_control/src/mutator.rs
@@ -426,6 +426,16 @@ where
     ) -> Result<()> {
         let wait_group = W::default();
 
+        // The first unfinalized block builds on the anchor, so the anchor's payload has to be
+        // replayed as well for that block to pass `BlockAction::DelayUntilPayload`. This has to be
+        // done before unfinalized blocks exist check below
+        if self.store.anchor().is_post_gloas() {
+            self.replay_persisted_execution_payload_envelope(
+                &wait_group,
+                self.store.anchor().block_root,
+            )?;
+        }
+
         let Some(last_block) = blocks.next_back().transpose()? else {
             self.finished_loading_from_storage = true;
             return Ok(());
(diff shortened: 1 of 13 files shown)


fork_choice_control/src/mutator.rs
```diff
@@ -465,13 +475,57 @@ where
...
+        let payload_envelope_identifier = envelope.as_ref().into();
+
+        let result = self.store.validate_execution_payload_envelope(
+            &envelope,
+            &origin,
+            &self.execution_engine,
+            NullVerifier,
+        );
+
+        // TODO: timeliness of a payload does not persisted, so we cannot know if it was seen before
```
povi: The only risk I see if restart and replay happens very quickly, the envelope of head block that previously was not timely is loaded as timely, thus potentially causing the client to build on full payload instead of empty. We should persist timeliness info as well.

fork_choice_control/src/storage.rs
```diff
@@ -1548,6 +1548,75 @@ mod tests {
...
+
+        envelope.message.payload.slot_number = slot;
+        envelope.message.beacon_block_root = block_root;
+
+        Arc::new(envelope)
+    }
+
+    #[test]
+    fn test_execution_payload_envelopes() -> Result<()> {
+        let database = Database::persistent(
```
povi: FYI, alternatively, you could use `Database::in_memory`. That would work too and wouldn't create real temporary files.

fork_choice_control/src/mutator.rs
povi: Also, change `does not persisted` to `is not persisted`. Normally I wouldn't mind, however, now there are bots that scan repos and create low value PRs that fix grammar, which is annoying

fork_choice_control/src/storage.rs
```diff
@@ -1548,6 +1548,75 @@ mod tests {
...
+        let mut envelope = SignedExecutionPayloadEnvelope::<Mainnet>::default();
+
+        envelope.message.payload.slot_number = slot;
+        envelope.message.beacon_block_root = block_root;
+
+        Arc::new(envelope)
+    }
+
+    #[test]
+    fn test_execution_payload_envelopes() -> Result<()> {
```
povi: What is this test? It tests that execution payload envelopes are pruned from the storage?

hangleang: technically the test is not needed, it has been removed during rebase 
7057551

fork_choice_control/src/mutator.rs
hangleang: 83cb00c store the timeliness only if the payload envelope arrived timely, so for existing key in old DB will be load as untimely, rather than assuming it's timely. what about the envelopes from RPC requests, during syncing we can't obtain the timeliness info from the responses

hangleang: it won't be matter for old data, but at least for the head payload envelope, it need the info to build the block correctly, or we could take PTC votes to decide which EL head to build on?

fork_choice_control/src/mutator.rs
```diff
@@ -438,11 +438,21 @@ where
 
         self.handle_tick(&wait_group, Tick::start_of_slot(head_slot))?;
 
+        // The first unfinalized block builds on the anchor, so the anchor's payload has to be
```
povi: One more thing caught my eye: if there is no unfinalized blocks to process, this code is unreachable (see the place above, where `finished_loading_from_storage = true` is set. Maybe add an anchor block envelope replay there too? Or make changes that it don't do an early exit if there is no unfinalized blocks

hangleang: move it before the unfinalized blocks exist check in rebased commit [745def4](https://github.com/grandinetech/grandine/pull/887/commits/745def433507856023e7f5b681764d8f4ab3cb5a)

PR #895, opened 2026-09-01
Prepare next slot execution payload again after PTC votes

hangleang:
At proposing slot, Grandine only call FCU to EL to prepare payload at `AggregateFourth`, which is the last tick in pre-Gloas. But for post-Gloas, the tick is just a tick before PTC votes, so EL won't know whether the payload reveal in time or not. This fix is to prepare the payload again at `PayloadAttestFourth`, at that point beacon node has received PTC votes from the committee, which allow it to allow it to build on correct branch based the votes.

### DIFF
--- a/fork_choice_control/src/mutator.rs
+++ b/fork_choice_control/src/mutator.rs
@@ -659,7 +659,11 @@ where
             }
         }
 
-        if tick.kind == TickKind::AggregateFourth {
+        // `TickKind::PayloadAttestFourth` is the last tick before the next slot in post-Gloas
+        if matches!(
+            tick.kind,
+            TickKind::AggregateFourth | TickKind::PayloadAttestFourth
+        ) {
             let store = &self.store;
 
             if let Some(state) = self.state_cache.existing_state_at_slot(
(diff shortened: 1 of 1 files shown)


povi (approved):

fork_choice_control/src/mutator.rs
```diff
@@ -659,7 +659,11 @@ where
             }
         }
 
-        if tick.kind == TickKind::AggregateFourth {
+        // `TickKind::PayloadAttestFourth` is the last tick before the next slot in post-Gloas
+        if matches!(
+            tick.kind,
+            TickKind::AggregateFourth | TickKind::PayloadAttestFourth
```
hangleang, line 665: prepare payload at `TickKind::AggregateFourth` in post-Gloas is to signal EL to build payload for the full branch if the payload is accepted, and at `TickKind::PayloadAttestFourth` is to check with PTC votes whether to reorg the payload if it reveal lately, and signal EL to use empty branch instead. An alternative approach is to prepare payload for full branch when payload is accepted, instead of unconditionally at `TickKind::AggregateFourth`

hangleang: so without the early prepare at the end of aggregate tick, EL will have just a small window (~0.75s before proposal) to freshly build new payload for full branch after PTC votes timely

Issue #860, opened 2026-08-21
Builder deposit cache


hangleang:
the current builder deposit signature cache reside only in `onboard_builders` function, so it only exist during the fork, we should move it to somewhere persistent between state transition, might be beacon state is the right place, also we need to handle some [edge cases and scenarios](https://github.com/jtraglia/kurtosis-devnets/tree/master/gloas-deposits) which haven't really tested end to end

nflaig:
> we need to handle some [edge cases and scenarios](https://github.com/jtraglia/kurtosis-devnets/tree/master/gloas-deposits)

keep in mind that you only need to handle s2,s3, and s5 from this list, all other scenarios no longer work with EIP-8282 as deposits are no longer interleaved between validators and builders and there can only be max 64 per block after gloas

PR #913, opened 2026-09-07
Update builder deposit cache so it handles some edge cases

povi:
Implements fixes for edge cases S2, S3 and S5, as per comment https://github.com/grandinetech/grandine/issues/860#issuecomment-5378419825


helper_functions/src/deposit_signatures.rs
```diff
@@ -90,6 +97,59 @@ pub fn is_valid_deposit_signature_cached(
     is_valid
 }
 
+/// Verifies the signatures of `deposits` in parallel, one result per deposit, in order.
+///
+/// [`is_valid_deposit_signature_cached`] verifies one signature at a time on a miss,
```
Tumas, line 102: is_valid_deposit_signature_cached has no production caller after the PR, and it is the only place the cache hit and miss counters are incremented, so both Prometheus metrics stay at zero forever.

povi: Added back cache hit & miss tracking

PR #919, opened 2026-09-16
compliance test fixes

povi:


hangleang:
There are more comp tests we skip https://github.com/grandinetech/grandine/blob/c5b25e8e3edd624009be0c3c54f5250c41a1244e/scripts/ci/consensus-spec-tests-coverage.rb#L14-L16

Should we handle them in this PR, or follow up with another PR?

povi:
> There are more comp tests we skip
> 
> https://github.com/grandinetech/grandine/blob/c5b25e8e3edd624009be0c3c54f5250c41a1244e/scripts/ci/consensus-spec-tests-coverage.rb#L14-L16
> 
> Should we handle them in this PR, or follow up with another PR?

Good catch! Added tests to run for other phases too. Everything passes except for these few failing tests. They all hit one problem: attestation target state processing for more than 2 epochs for non-canonical chains. This one change warrants it's own PR.

fork_choice_control/src/compliance_tests.rs
```diff
@@ -112,23 +112,73 @@ struct ViableHeadRootAndWeight {
 }
 
 // TODO: fix remaning failing compliance tests
```
hangleang: typo here

hangleang (approved):

Issue #713, opened 2026-06-01
[Glamsterdam] enable again wait for fully validated head for validator duties


povi:
Needs a fix for `optimistic` head status first: https://github.com/grandinetech/grandine/issues/712

eth2353:
Hey povi, I just tried running Vero against Grandine on a local devnet and noticed the `head_v2` SSE events are emitted with `execution_optimistic=True`. I assume that is related to this issue but if it's not, let me know and I'm happy to create a new one for that.

povi:
Hey, yes, that is related to this issue

PR #928, opened 2026-09-18
Invalidate Gloas payload statuses when a payload is invalid

hangleang:
`invalidate_block_and_descendant_payloads` only sets `payload_status` to `Invalid`, rejecting the beacon block [already guarded](https://github.com/grandinetech/grandine/blob/34da987dc6e1eec7be6afb6fb3c6edf317f1c32c/fork_choice_control/src/mutator.rs#L756) on post-Gloas block


povi:
This fix is not correct. Since Gloas, execution payload is separated from the block. And it's not block that gets invalidated, it's execution payload envelope, meaning only the FULL block is invalid. The EMPTY block still can be a part of the canonical chain.

If the beacon node invalidates the block when EL says that payload is invalid, it could get stuck forever if the network decides to go with the EMPTY block.

There are a collection of inconsistencies if the fork choice, and I'm currently working on a set of fixes for that (issues #712 && #713).

hangleang:
> If the beacon node invalidates the block when EL says that payload is invalid, it could get stuck forever if the network decides to go with the EMPTY block.

how should CL respond when EL says a payload is invalid? should CL mark the invalid payload by its block hash, rather than beacon block root?

povi:
> > If the beacon node invalidates the block when EL says that payload is invalid, it could get stuck forever if the network decides to go with the EMPTY block.
> 
> how should CL respond when EL says a payload is invalid? should CL mark the invalid payload by its block hash, rather than beacon block root?

It should mark the chain link payload status as invalid, just like you did. However, the necessary missing part is how the invalid payload status is handled. Right now, if the block with invalid payload is invalid, block is non-viable. We need to change head selection rules to allow empty block in head selection. This is one thing I'm working on. And without this part, if your PR is merged, there is a risk of beacon node stalling.

hangleang:
> And without this part, if your PR is merged, there is a risk of beacon node stalling.

I see now, I will undo that part and keep only the genesis fix

